PR #63, opened 2026-08-13
build: the toolchain runs under Linux pwsh -- the QEMU fallback learns the codex-vm file contract

showell:
## What this is

I have been exploring Codex from an outside git clone on Linux (WSL2), a machine with no codex-vm.exe. This PR makes the stock toolchain work there: `pwsh build/compile.ps1 -Src <src> -Out <out> -Log <log> -Kernel seed/Codex.cdx` boots the seed under stock QEMU 6.2 with TCG and compiles a small test in about 10 seconds wall including boot.

Everything is changed via the generators (vmconfigScript.codex, CompileScript.codex) and regenerated, per the vm-config.ps1 header policy: the shipped .ps1 files are the regenerated output, both re-emit at zero drift by check-generated-scripts' own trim-compare, parse clean, and carry no unhandled-node stubs. The baseline is untouched.

## The two boot gaps (likely relevant beyond Linux)

Two things codex-vm does pre-boot that QEMU does not:

1. codex-vm writes the guest RAM size at phys 0xFE8 ("dynamic RSP", ram-size-addr in Emit/X86_64Boot.codex). Without that cell the boot stub computes RSP from zeroed memory and triple-faults before the first serial byte. The QEMU path now seeds it with `-device loader,addr=0xfe8,data=<ram-bytes>,data-len=4` (data-len 4 because QEMU's generic loader asserts data_len < 8, and -m stays under 4 GB).
2. The default qemu64 CPU model lacks features the seed's code uses, which is the same silent triple-fault. `-cpu max` boots.

I suspect the repo's other QEMU paths (gdb-watchpoint, boot-arm64's x86 sibling) share gap 1. I cannot test Windows from here, so those are left untouched.

## What changed

**vm-config.ps1**
- A missing codex-vm no longer throws at dot-source time. That throw made the QEMU fallback below it unreachable on any machine that never had codex-vm. The hard failure is reserved for having neither host.
- QEMU discovery gains the Linux side: PATH lookup and a QEMU_BIN env var, with QEMU_BIN_WHPX still honoured. The accelerator comes from CODEX_ACCEL, defaulting to tcg. Measured here, the nested-virt vmexit tax on byte-wise serial I/O makes KVM slower than TCG for small compiles (18-63s against ~11s), so kvm is the opt-in.
- kernel-irqchip=off is retained for WHPX/TCG and dropped under KVM, where the userspace-APIC path it forces dies before READY.
- New `Invoke-VmCompileFallback`: one compile over the serial wire honouring codex-vm's -input/-output file contract, so compile.ps1's parsing is host-agnostic. The input is paced in small flushed chunks: a single large burst intermittently lost bytes on the way into the guest, and a lost EOT hangs both sides (codex-vm is immune because it preloads the ring). The read stops when the SIZE payload is complete plus a short trailer drain, or when a CODEGEN/!EXC report goes quiet, rather than on idle alone.

**compile.ps1**
- Dispatches on whether codex-vm exists; the codex-vm arm is unchanged.
- Two path joins made cross-platform (the build-output default kernel and the seed drift-note path).

Windows behavior is intended to be unchanged: the whpx and codex-vm arms carry the same argument lists as before, and the platform additions are gated on not-Windows.

## Verified (WSL2, QEMU 6.2 TCG, pwsh 7.5, seed 527C2C75)

- codex/test/act-let-scope.codex compiles in ~10s wall including boot, the .map sidecar is captured, and the binary boots and matches its .expected byte for byte
- a deliberately broken source exits 4 in ~10s with diagnostics remapped to the caller's file and line
- both generators compile (~42s each) and re-emit the shipped scripts at zero drift
- `-WindowStyle` is guarded: it throws on non-Windows editions of pwsh

## Not covered

- The Windows arms are preserved by construction but untestable from this machine.
- test-run.ps1 and the other direct codex-vm callers still need codex-vm; compile.ps1 was the target here. Happy to give test-run.ps1 the same treatment if wanted, or to reshape any of this to taste.


damiant3:
Thank you, @showell. This is the first contribution to the project from outside the fleet, and it is now public.

**Landed in Update 41**, commit [`baee98b`](https://github.com/damiant3/NewRepository/commit/baee98bd) on `master`. The design and the first working version are yours; you can check the claim against the tree rather than take our word for it.

Because the project is Perforce-primary with GitHub and GitLab as downstream mirrors, your commits could not be merged where they were opened -- the work was carried into the depot and comes back out through the release push. That is a shape git tooling does not expect, and the reason this PR stayed open until now: until the push went out, "this is in" was not true on any tree you could see.

**Four tweaks were made on top, each measured on Windows 11 rather than argued:**

1. **`CODEX_VM_HOST=qemu` forces the fallback on a box that has codex-vm.** Without it the QEMU path cannot be exercised from Windows except by hiding the binary, and that is how it stayed dead: measured, the Windows QEMU fallback in `Start-VmRun` had never booted on main.
2. **The two boot gaps track the accelerator, not the OS.** Under whpx, `-cpu max` wedges QEMU 11.0.0 (fixed in 11.1.0; the guard stays because 11.0.0 still needs it and 11.1.0 only makes it redundant). The 0xFE8 RAM-size cell is required on every accelerator on both versions.
3. **QEMU discovery prefers a side-loaded versioned install, newest first.**
4. **The fallback read loop breaks when the guest process has exited.** A read timeout and an abortive socket close arrive identically as an exception, so a dead guest spun the loop hot until the full deadline.

You also named the 0xFE8 private ABI as a compromise, and you are right: nothing in the compiler ever writes that cell, and the multiboot header at `X86_64IO.codex:4` does not set `MEMORY_INFO`, so the guest never asks the loader the question whose answer it is being handed. Retiring that cheat is recorded as open work in Update 42, not signed off as done.

The durable account is [`docs/PM/Done/GitHubUpdates/GitHubUpdate41.md`](https://github.com/damiant3/NewRepository/blob/master/docs/PM/Done/GitHubUpdates/GitHubUpdate41.md) -- your PR opens the report.

Closing as landed. Thanks again.

PR #64, opened 2026-08-14
Zig plug warmup

showell:


damiant3:
cool.  ill have opus jump on this right away.

showell:
Just a heads up that we are still actively working on this PR.

damiant3:
Both of your batches are public as of `851c1aa2` (Update 43). Thank you -- this
was the most productive thing anyone has pointed at the compiler this cycle,
and the reason is the one you gave: you were building something that had to
agree with us, and the defects fell out of the attempt.

What landed, and what we changed on top:

- **`net-recv-raw`** went in as you wrote it. You offered it uncompiled and
  said so; we compiled and gated it. The only edit was trimming the prose block
  above the helper to the two facts a reader cannot derive from the code beside
  it (REP INSW moves whole words, NE2000 pads to 60 bytes but never to even) --
  a house rule about commentary, not a disagreement.
- **The ARP reply** went in unchanged. We added `arp-reply-test`, whose second
  arm is a control: a request for a different IP must still answer
  `arp-learned` and queue nothing, so a function that replies to everything
  cannot pass.
- **`bytes-to-text`**: rather than fix it in 40 plugs we moved one linear copy
  into `plugs/common/PlugTypes.codex`, which the plug builder already bundles
  into every transpiler plug, and deleted all 42 local copies.
- **`deck-record`**: we took your guard side. Your `probe-deck-init` already
  falsified the other contract. The discriminator turned out to be in the IR
  as `IRDef.chapter-slug`, so the intrinsic is now the `deck-record` defined
  beside `init-phase-allocator`; in a plug bundle it is the plain identity
  `PlugTypes` always advertised. Your six probes ran both ways: five arms move
  to correct, both controls stay put.
- **Finding 5** became **CDX2096 UnreachableMatchArm**, generalised past the
  instance to any arm following an *unguarded* catch-all. Error rather than
  warning, on the evidence you started: the 270-unit sweep finds no other case.
- **Finding 4** became a cite on `TypeChecker`.

Two corrections, both ours to make and neither changing your conclusions.
`bytes-to-text` was quadratic in **39 of 42** definition sites once your own zig
fix is counted, not 42 of 44. And `repro-crash` does not page fault on our
current seed -- it answers wrongly. You had already said the severity was
allocation-pattern luck, and it is. Both are recorded next to your claims in the
Update 43 report rather than edited into them.

**What we could not check, said plainly:** your central result, that the emitted
zig runs byte-identical to bare metal on the lexer and the whole parser. There
is no zig toolchain on our build box and zig is not wired into
`plug-oracle-test.ps1`, so that remains your measurement on your machine. It is
filed as an open gap on our side, not a doubt about yours.

You also found a class rather than three defects. All three compiler findings
came out of bundling a *subset* of the compiler into one unit, which our
glob-assembled build hides completely. We now have
`build/check-subset-cites.ps1`, which compiles every chapter with only what it
cites; on its first run it found `BootPaint` borrowing `to-unicode` from
Foreword CCE while citing nothing at all. That one is ours, and we would not
have gone looking without your PR.

The full write-up, including your own account quoted in full, is in
`docs/PM/Active/GitHubUpdates/GitHubUpdate43.md` at the commit above.

**Leaving this open** rather than closing it, since you said the work is still
in progress and the five newer zig commits are yours to finish.

damiant3:
Thank you, @showell. This one is absorbed, and it earned its keep before a line of the zig work was judged: the first batch was five separate fixes, each gated on its own CL, and one was a live data-corruption bug none of our own tests had caught.

**Absorbed in Update 43** (Perforce main 15184 and 15190). What landed:

1. **`net-recv-raw` rounded the `REP INSW` word count down**, so an odd-length frame dropped its last byte and the caller read a byte of the *previous* frame at that offset. A real byte from the same stream, so it parsed and was never diagnosed. Your measurement: 11 corrupt transfers in 15 on a 191 KB payload.
2. **The stack learned ARP senders and never answered requests.** `arp-build-reply` had sat in `Ethernet.codex` with no caller, invisible here because codex-vm's NAT knows the guest MAC by construction and never asks, and it killed a plug's connect under QEMU's slirp. The new `arp-reply-test` carries its own control so a function that replied to everything could not pass.
3. **`bytes-to-text` was O(n^2) in 39 of 42 sites**, collapsed to one linear definition in `plugs/common/PlugTypes.codex`. On a 2.19 MB IR the quadratic form went out of memory with no output where the linear one emitted in 42 s.
4. **The x86-64 deck-record intercept fired on the name**, so every plug kernel ran `__deck-enter`/`__deck-exit` against an allocator it never initialized. It fires on the defining chapter now.
5. Your second batch produced the cycle's one new diagnostic: two identical `is otherwise` arms in one `when` in `TypeCheckerInference`, the second unreachable. **CDX2096 UnreachableMatchArm** now refuses any arm following an unguarded catch-all.

The ten commits of zig plug work came in with them and are recorded honestly as **our measurement, not yours**: there is no zig toolchain on the build box, so the byte-identical claim could be reproduced here only through codex-vm as the truth arm. That standing caveat is the one thing this project cannot lift for you unaided, and it is stated plainly in Update 43.

Because the project is Perforce-primary with GitHub and GitLab as downstream mirrors, the commits could not be merged where they were opened; the work was carried into the depot and comes back out through the release push. The durable account is [`docs/PM/Active/GitHubUpdates/GitHubUpdate43.md`](https://github.com/damiant3/NewRepository/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate43.md), on the public mirror now, where your PR opens the report.

**#65 is where the work continues** and is open: the fourteen commits absorbed here are gone from it, and what remains is the eight-rung ladder replayed onto the Update 43 seed. Closing this one as absorbed. Thanks again, Steve.

PR #65, opened 2026-08-16
zig plug: the eight-rung ladder, rebased onto Update 43

showell:
Continues the zig plug work from PR #64, rebased onto Update 43. The
fourteen commits you absorbed are gone from here; what remains is the work
that came after, replayed onto the new seed and re-verified against it.

## What this is for

Making the zig plug reliable enough to stand in for the bare-metal front
end in a mode that emits text, and using the effort of getting there to
find defects in Codex. The port is the vehicle.

## Measurement

Eight rungs, each bundling real compiler chapters into a subject, compiled
two ways -- seed on bare metal as truth, and through the zig plug -- and
required to agree byte for byte:

    lex -> parse -> desugar -> scope -> check -> lower -> text -> pingpong

All eight pass on seed F3722EAC. `text` emits canonical codex source out
of the IR through the plug; `pingpong` compiles that emitted source again
and requires the second text to equal the first. The fixed point holds
through the plug.

Re-banking against your new seed produced one result worth reporting: every
truth file came back byte-identical to the pre-Update-42 bank, across two
updates that touched 200 files including `Types/Unifier.codex`,
`IR/Lowering.codex`, `Types/TypeCheckerInference.codex` and
`Core/PhaseAllocator.codex`, all of which are inside these subjects. The IR
and the emitted zig moved. The observable behaviour did not.

## Findings

`deck-record-repro/README.md` now marks 1 through 5 CLOSED with what
landed, and keeps them only as a record. Three are open: `lift-lambdas`
sitting behind the plug wire, the absence of any `IRExpr` map or fold, and
`passes=text-plug` changing the type vocabulary without saying so.

Finding 4 is worth a second look. It closed as `check-subset-cites.ps1`
rather than as the one-line cite it asked for, which is a gate for the
whole class instead of an answer to the instance. Two of our own
accommodations broke when it and finding 5 landed -- a bundle that named
`Capability` outright, and an emitter check that dropped a duplicate
catch-all arm. Both are removed here. An accommodation is a record of a
defect held somewhere that cannot know when the defect goes away.

## The standing caveat

Update 43 records it fairly: the central claim here is our measurement, not
yours, because there is no zig toolchain on the build box. Nothing in this
PR changes that, and we would like to. Related and not included here: the
zig plug is absent from `build/plug-oracle-test.ps1` entirely, so its
status there is "not listed" rather than "unverified". Running that
subject through the zig plug by hand finds three real defects in our plug
-- a missing `int-mod` emitter, a record literal emitted without the
parentheses zig needs before a field access, and bounded-field clamping
not emitted at all. We would rather fix those and then add the arm than
add an arm we know is red. Say the word if you would prefer it listed and
failing.

Also worth knowing: `plug-oracle-test.ps1` runs `tools\codex-vm.exe`
directly for the truth arm rather than going through `compile.ps1`, so it
cannot run on a non-Windows box even where `compile.ps1` can.

### DIFF
--- a/codex/plugs/zig/ZigEmitter.codex
+++ b/codex/plugs/zig/ZigEmitter.codex
@@ -177,7 +177,7 @@ Section: Type Mapping
     is FunTy (p) (fnrow) (r) -> emit-zig-fn-type ty
     is EffectfulTy (effs) (scopes) (inner) -> emit-zig-type inner
     is ForAllEff (c) (inner) -> emit-zig-type inner
-    is RecordTy (n) (a) (f) -> zig-sanitize (n.value)
+    is RecordTy (n) (a) (f) -> zig-sanitize (n.value) & zig-type-args a
     is SumTy (n) (a) (c) -> zig-sanitize (n.value) & zig-type-args a
     is ConstructedTy (n) (a) -> zig-sanitize (n.value) & zig-type-args a
     is TypeCon (n) -> zig-sanitize (n.value)
@@ -395,6 +395,13 @@ Section: Constructor Map
  body has to follow. A const alias does not avoid the rule; it reaches
  into the nested struct's function too.
 
+ bound-names lists the codex names that are runtime locals of the zig
+ function being emitted -- its parameters, its let bindings, its match
+ binders. Top-level definitions are not in it and never need capturing.
+ A lambda intersects this list with the names its body mentions, and what
+ falls out is exactly what its environment has to carry, because zig lets
+ a nested function reach a comptime constant but not a runtime value.
+
   ZigCtx = record {
    arities : List ArityEntry,
    ctors : List ZigCtorEntry,
@@ -405,7 +412,8 @@ Section: Constructor Map
    renamed-to : List Text,
    tvar-ids : List (Integer between 0 and 4294967295),
    tvar-types : List CodexType,
-   expected : CodexType
+   expected : CodexType,
+   bound-names : List Text
   }
 
  An empty list literal reaches the IR without a concrete element type,
@@ -479,7 +487,7 @@ Section: Constructor Map
    in if di < 0 then VoidTy
    else let d = list-at (ctx.irdefs) di
    in if i >= list-length (d.params) then VoidTy
-      else zig-subst-all ((list-at (d.params) i).type-val) ctx 0
+      else zig-subst-ctx ((list-at (d.params) i).type-val) ctx
 
   zig-param-ll-elem : ZigCtx, Text, Integer -> Text
   zig-param-ll-elem (ctx) (fname) (argi) =
@@ -491,7 +499,7 @@ Section: Constructor Map
    else let d = list-at defs i
    in if d.name == fname then
     (if argi >= list-length (d.params) then ""
-     else let pty = zig-subst-all ((list-at (d.params) argi).type-val) ctx 0
+     else let pty = zig-subst-ctx ((list-at (d.params) argi).type-val) ctx
      in when pty
       is ListTy (elem) -> emit-zig-type elem
       is LinkedListTy (elem) -> emit-zig-type elem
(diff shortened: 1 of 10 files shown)


damiant3:
Carried into the depot, @showell. Thank you.

The emitter went in as a clean file-replace: the depot copy of `ZigEmitter.codex` was byte-identical to what you branched from (line endings aside), so there was no merge to do, which is the nicest possible shape for a rebased contribution. The plug rebuilds here and emits for a real subject.

The eight-rung ladder is recorded, and so is the one result you flagged as worth reporting: every truth file coming back byte-identical across two updates that moved `Unifier.codex`, `Lowering.codex`, `TypeCheckerInference.codex` and `PhaseAllocator.codex` is exactly the property a second implementation held to byte-agreement is supposed to catch, and it caught nothing because nothing was wrong. That is the instrument working.

The standing caveat is unchanged and stated plainly on our side: there is no zig toolchain on the build box, so `text` and `pingpong` agreeing byte-for-byte is your measurement through the plug, not ours through a zig compiler. We would rather fix the three defects you already found by hand-running the parser subject -- the missing `int-mod` emitter, the record literal emitted without the parens zig needs before a field access, and bounded-field clamping not emitted -- and *then* add the `plug-oracle-test` arm, than add an arm we know is red. All three are recorded as open in `plugs-backlog` 1.13.

`deck-record-repro/` stayed on your branch: it is your repro workspace and has no home in the depot, and that is the right place for it.

Because the project is Perforce-primary with GitHub as a downstream mirror, the work is in the depot now but not yet on any `master` tree you can see; **this PR stays open until the release push goes out**, the same shape as #63. When it does, the durable account is in `docs/PM/Active/GitHubUpdates/GitHubUpdate44.md`. Good progress, Steve. Thanks again.

damiant3:
Good question, and no -- that sentence was about our side of the wiring, not a fence around the fixes. It clearly read the wrong way, so let me be exact.

We will not add the zig arm to `plug-oracle-test` while it is red, and the arm is red until those three defects are fixed. So the sequencing on our side is: fix (or receive) the three, then wire the arm. That is all it meant. It was never "leave these alone" -- they are your finds, and a contributor fixing the defects a port surfaces is exactly the loop we want. Please do have your Claude carry on.

Your proposed shape is the right one, and better than ours: **one commit per fix in the next PR.** That lets us take each independently and drop any that collide with something we land in the meantime, with no rebase pain for you, and it keeps each fix reviewable on its own -- which matters here precisely because we cannot compile the emitted zig on this box.

Two small things so a fix lands clean on our side:

- Carry your measurement in the commit the way the ladder does: a subject through the plug, byte-identical to bare metal, named in the message. We absorb it as your measurement, since there is still no zig toolchain here to reproduce it (the standing caveat, unchanged).
- On `int-mod` specifically: note whether you emit floor or truncating. Bare-metal `int-mod` is floor (non-negative remainder); `math-mod` is truncating. The plug today emits `@rem` for `IrRemInt`, which is truncating, so `int-mod` needs the floor correction (zig's `@mod`), not just a missing-name fix. Easy to get subtly wrong in the direction that passes small tests and fails on negatives.

We have the three recorded as open on our side in `plugs-backlog` 1.13. If your Claude is already on them we will hold off duplicating and take yours; say the word if you would rather split them. Thanks, Steve.

showell:
See Steve's chat for reply.

damiant3:
Thank you, Steve. This is landed and public as of Update 44, commit ff1d299b on `master` (Perforce main 15700).

What went in: your eight-rung ladder replaced `codex/plugs/zig/ZigEmitter.codex` wholesale (the depot was byte-identical to your base, so the replace was clean, carried in as main 15597), and `deck-record-repro/README.md`'s closed/open split is recorded as you wrote it. Your work continues to be credited by name in the report.

What changed on our side since you opened this: zig 0.16.0 is now installed on the build box, so the standing caveat from Update 43 is lifted. The plug was verified here first-hand (main 15634): the core is byte-identical to bare metal on a subject with arithmetic, records, recursion and text concat, and your three self-reported defects reproduce exactly (`int-mod` missing, and note `@rem` is the wrong fix on negative dividends, it needs `@mod` plus the sign correction; the unparenthesized record literal before a field access; bounded-field clamping not emitted). We found a fourth: a list literal is emitted with element type `void` (`&[_]void{ 10, 20, 30, 40 }`), so `list-length` on a literal does not compile. All four are tracked as `codex/plugs/plugs-backlog.md` 1.13 and the oracle arm stays unwired until they land, per your own position in this thread.

Report: https://github.com/damiant3/NewRepository/blob/ff1d299b750a9d7710350d05c755623206edf90b/docs/PM/Active/GitHubUpdates/GitHubUpdate44.md (section "The zig plug is verified here now, and PR 65 is in").

Closing this PR referencing ff1d299b; the mirror is downstream of Perforce, so the merge happens there rather than here. Further zig work is welcome as a new PR against master.

damiant3:
Closed as landed in ff1d299b (Update 44). See the comment above for what was verified and what is open.

PR #66, opened 2026-08-16
zig plug: int-mod, record-literal parens, and the bounded-field clamp

showell:
Closes the three defects self-reported on PR 65 and recorded in plugs-backlog 1.13: the missing int-mod emitter, record literals rejected before field access, and bounded-field clamping not emitted. One commit per fix, as requested, so each can be taken or dropped independently; the messages carry the reasoning and the measurement.

Stacked on zig-plug-ladder (PR 65, red 15595 / main 15597), so only the top three commits are new here; the rest becomes part of master at the next release push.

Measurement, ours by the standing caveat, taken at the batch head: codex/test/plug-oracle-arith.codex seed-compiled to IR-CCE, through the plug, zig-run -- 20 of 20 values byte-identical to the bare-metal run of the same subject, and that truth matches the checked-in .expected. The eight ladder rungs re-swept green after the batch. The subject exercises all three defects, so its emission first compiles at the head; the intermediate commits say so rather than claiming a measurement they cannot have.

Two notes from the work:

- int-mod is Euclidean rather than floor: the subject's prose says so and the bare-metal truth confirms it on md 7 -3 (bare metal 1, zig @mod would answer -2). Details and the full disagreement table are in the int-mod commit message.
- Bare metal clamps a bounded field at every store site (record literal, __record-set, field store all reach emit-clamping-checked-store); the C# plug clamps only at construction, so its other two store paths carry the same species of silent disagreement the arith subject's prose predicts. Noted rather than fixed, since that call is yours.

Per your comment, the plug-oracle-test zig arm stays yours to wire once these land.


showell:
All but the last three of these commits (as of this writing) were on #65 and already addressed there.

damiant3:
A correction I owe first: my note upthread said `int-mod` is floor and pointed at zig's `@mod`. That is wrong, and your commit has it right. `int-mod` is Euclidean, neither floor nor truncating, and I can confirm it from the same box you cannot run on: on the arith truth, `@rem` fails the negative-dividend lines (`md -7 3` gives -1 where truth is 2) and `@mod` would fail the negative-divisor lines (`md 7 -3` gives -2 where truth is 1); only Euclidean matches all five `md` lines. My "non-negative remainder" was the right property with the wrong name attached. Thank you for the catch, and for spelling out the distinguishing lines in the commit message.

The measurement shape is exactly what lets us take it: `plug-oracle-arith` through the plug, 20/20 byte-identical, the ladder re-swept, all flagged as your measurement under the standing caveat. Clean.

On the three-way split: **do not bother.** One verified commit is fine on our side. We carry a PR into the depot by hand regardless, so if we want each fix as its own change we will split the hunks at absorption; there is no reason for you to redo the commit for our internal CL hygiene. Keep it whole.

One thing this PR will not cover, found here while verifying (zig 0.16.0 is on this box now, so this is our measurement too): a list literal `[10, 20, 30, 40]` is emitted with element type `void` (`&[_]void{ 10, 20, 30, 40 }`), so `list-length` on a literal will not compile (`comptime_int` vs `void`). The arith subject has no list literal, so your 20/20 is real and this is simply the next one along: element-type inference on a list literal is missing, likely the same empty-list-element-type gap one context over. Recorded on our side in `plugs-backlog` 1.13, as a follow-up for whenever you are in that code, no urgency.

And noted: you keep `plug-oracle-test.ps1` as your step after the fixes land, so we leave the harness alone. That matches what we recorded here. Good work, Steve.

damiant3:
Thank you, Steve. This is now public in **Release 45**, commit [`c5b95be0`](https://github.com/damiant3/NewRepository/commit/c5b95be02f77a0b95a9ebca94677c107e02cceef) on `master`, and the writeup is in [GitHubUpdate45.md](https://github.com/damiant3/NewRepository/blob/c5b95be02f77a0b95a9ebca94677c107e02cceef/docs/PM/Active/GitHubUpdates/GitHubUpdate45.md) under "PR 66 carried in".

Carried your `ZigEmitter.codex` wholesale onto main. It takes the zig oracle arm from red to green and every wired arm is now 33 of 33, the first time that has been true: `build/plug-oracle-test.ps1` is 5 passed 0 failed, python, javascript, zig, wasm and csharp each 33 of 33. The one thing added on top was a `list-snoc` registration beside your `list-push`, pointing at the same `cx_ll_push`. Without it the shared oracle subject emits `@compileError("zig plug: no emitter for list-snoc")` and does not compile.

Rung 13 we could not reproduce on Windows: the plug guest raises `OUT OF MEMORY` part way through the whole-compiler emission, stopping between 534,800 and 547,400 bytes across five identical runs, every length a multiple of `net-mss` 1400, and it is not the guest's memory size (3 GB and 12 GB stop in the same band). We got as far as your `bundle_whole.ps1` building the subject here at 54,856 lines, compiling at `-Decks 172`, running on bare metal, and emitting 13,488,840 bytes of IR in 118 s. Your 16,874-line result stands as your measurement on your harness; the process form is Linux-only and is recorded the same way.

Your `run.ps1` is the one file we did not take: main's passes `-Passes 'text-plug'`, which a source plug needs, and yours does not.

**Re-checked before closing, because a close should be checkable.** Our `codex/plugs/zig/ZigEmitter.codex` is byte-equal to this PR's current head `fe6afa37` plus that single `list-snoc` line, so "wholesale" is still literally true against your newest commits and not just the head we absorbed. Of the 151 paths the PR touches, that is the one we took: `zig-ladder/` stays on your branch as PR 65 set the precedent, and the `deck-record-repro/` probes stay with it.

**What we had missed, and thank you for it.** Your `deck-record-repro/README.md` is a findings register, and findings 6 through 10 plus candidate 11 are about our compiler rather than about zig. They were in no register on our side, so closing this PR would have closed the only copy. They are now recorded as row 1.27 of `codex/plugs/plugs-backlog.md`, attributed to you and this PR, with your probes named as what runs them. Two we verified here rather than relay:

- **Finding 6** holds exactly as you state it. `lift-lambdas` has one caller, `opening.codex:833`, inside the `cdx-chapter` path; `emit-ir-cce` runs `compile-frontend-ir` and the named pipeline and never reaches it. So the plug wire does carry lambdas closing over enclosing locals, nothing in the interface says so, and your point that lifted IR would have arrived as a partial application the plug already handled is the part that stings. The free-variable analysis you had to write was our omission, not yours.
- **Finding 9 reproduces, both directions.** Against the seed shipped in this release, `f : PhantomSig -> Integer` with `PhantomSig` defined nowhere compiles at exit 0 with no diagnostic, while the same shape with an undefined VALUE name is refused `CDX3002: Undefined name`. Type names in annotations are not existence-checked the way value names are. That is the sharpest of the six and it is ours.

Findings 7, 8, 10 and 11 are recorded as your measurements and labelled that way; we have not re-run them. We noted that you flagged 11 as a hypothesis rather than a diagnosis, and it is filed that way.

Closing this as landed in `c5b95be0`, per the note in `docs/Agents/PerforceProcess.md` section 7: the mirror is downstream of Perforce, so a PR is never merged where it was opened, and the work is re-applied in the depot with credit instead. Nothing about that is a comment on the contribution. Thanks again, and the zig plug is better than it was because you kept going past the point where the defects were someone else's.

damiant3:
Landed in Release 45 as c5b95be02f77a0b95a9ebca94677c107e02cceef. Closing per the note above: the mirror is downstream of Perforce, so the work was re-applied in the depot with credit rather than merged here.

PR #67, opened 2026-08-17
zig plug: the ladder on seed 270227BE, a hosted compiler, and a latent record-layout inconsistency

showell:
Our commits replayed onto Update 45. **No depot files change** — `git diff
upstream/master HEAD -- codex/ build/ docs/ seed/` is empty. PR 66 carried the
emitter work in whole, and this branch takes the depot's `ZigEmitter.codex`
verbatim including the `list-snoc` row added there.

What this adds is `zig-ladder/` (the harness) and `deck-record-repro/` (the
findings record). Take either, both, or neither.

## The one thing worth your time

**`emit-record` has two layouts and they disagree** (X86_64Compound.codex:1751).
It picks by whether the record's type resolved:

| | ranked by | offsets |
|---|---|---|
| type resolves — `build-cce-byte-offsets` | `field-sort-key` = `width-prefix(field-byte-width) & name` | cumulative field widths |
| type does not — `emit-store-record-fields-by-list` | the raw field name | `rank * 8` |

Every reader uses the first rule unconditionally — `cce-byte-offset-and-type` /
`accumulate-offset-width-sort`, in both `emit-field-access` and
`emit-record-set-builtin`. The two coincide only when every field is eight bytes
wide, because only then does the width prefix stop affecting the order and the
uniform slot match the real width.

`deck-record-repro/probe-record-layout.codex` is the smallest record where they
part:

    Box = record { flag : Boolean, items : List Integer }

    resolved:  items @ 0   flag @ 8    (size 9)
    fallback:  flag  @ 0   items @ 8   (size 16)

Built through the fallback, `flag` occupies the slot `items` is read from, so
`b.items` evaluates to the boolean — 0 or 1 — and `list-length` of that reads
`[0 - 8]`, faulting with CR2 at the top of the address space.

**Hedged properly: we could not reach that path through `opening.codex`.** The
probe compiles and prints `3` through the normal driver, because the type
resolves and the first layout is used. We reached the second only with a harness
that skipped the RESOLVE phase, which was our defect. So this is a latent
inconsistency visible by inspection, not a demonstrated bug, and it may be
unreachable today.

Why I would still close it: given the same unresolved type, `emit-field-access`
records an error and emits `ud2` — it refuses, and says so. `emit-record`
silently picks a layout no reader shares. Either make the `is otherwise` branch
an error like its neighbour, or have it use `cce-byte-offset-and-type` so the
two cannot drift. The first is probably right if the branch is dead, and it
turns a future silent-corruption bug into a diagnostic.

## Finding 11 is withdrawn

The earlier version of this description reported a `DiagnosticBag` with 72
errors and an empty diagnostics list, and asked why the two arms disagreed about
the number. **That was our bug.** The harness stands in for `opening.codex` and
skipped the RESOLVE phase (`build-type-def-map`, the deep-resolved binding set,
`rewrite-ir-defs`), so records declared in a subject had no entry in the
emitter's `st.type-defs`. `deck-record-repro/README.md` records it as withdrawn
with the cause. Nothing else in this PR depended on it.

## The ladder, for context

Fourteen rungs. Each bundles part of the compiler into one subject with a
generated harness, compiles it twice — the seed on bare metal as truth, and
through the zig plug into a zig program — and requires byte-identical output. It
ends at:

- **`whole`** — the whole compiler minus `opening.codex`, 16,874 lines of
  transpiled zig, byte-identical output including the same inlining decision
  (dropping the inline passes moves `scale-by-four` 22 → 29 bytes; the zig side
  emits 22).
- **`zigc`** — the same compiler with stdin/stdout instead of a Text literal.
  Builds in 4s, compiles a program in ~3s, and for one subject its output was
  byte-identical to the seed's and booted.

`zig-ladder/README.md` describes all of it, including what the check does and
does not prove — the zig arm's code generation has independent lineage (and goes
through zig's own x86-64 back end, not LLVM, measured), but the seed still
compiles the subject to the IR both arms rest on, so it is a DDC check over the
part of the pipeline after the IR rather than a complete witness.

## Current state, stated plainly

Thirteen of the fourteen rungs pass on seed 270227BE. `clamp` — the one rung
whose subject deliberately does not compile cleanly — page-faults on the bare
metal arm inside `__list_snoc`, called from `bag-add`, with a null list. That is
under investigation and is **not** the layout issue above; it reproduces with
and without the RESOLVE phase and with and without our parse scan, and it did
not occur on the previous seed.

**The banked truths are mid-flight.** A harness change today invalidated them,
and they are being re-run. Numbers from this branch should not be quoted against
this seed until that finishes.


showell:
Correction on finding 11, before it costs anyone time.

**The claim in the description is not currently reproducible and may be my own
measurement error.** The reading was `emit-errors 72` with an empty diagnostics
list. At the time the harness printed the bag's *count* but not its *contents*,
and my diagnostic walker printed nothing — which I read as "the list is empty"
when it could equally have meant "the walker is broken." I had no control to
tell those apart.

The harness now prints `emit-diags`, the length of the same list, as a number
beside the count. With that in place the same rung on seed 270227BE reports
`emit-errors 1` and `emit-diags 8`, and the walker prints all eight. So the
walker works, and the specific 72-with-nothing pairing has not reproduced. I am
re-measuring under the original conditions and will either restate the finding
with evidence or withdraw it.

What that investigation *did* turn up is better, and it is not a defect in the
depot:

    diag 2000 sev 0 emit-field-access: unresolved type for field 'g' addr=0 tag=0

`emit-field-access` handles `RecordTy` and sends everything else to
`is otherwise`, which records that error and emits `[15, 11]` — `ud2` — where
the load belonged. That is the right thing to do, and `emit-binary-tail` then
refuses to write a binary when the bag has errors.

Our harness did not have that gate. It printed the bytes anyway, so we ran a
binary the compiler had already rejected and spent a while being surprised that
it trapped with `!EXC=06` at `gauge`, whose field is
`Integer between -100 and 100 clamping`. The trap was the compiler being
careful; the bug was ours. The harness now gates on `bag-has-errors` the way the
driver does.

The open question worth your eyes is narrower than what I originally wrote: why
that field's record type arrives unresolved at `emit-field-access` at all, given
the same source compiles cleanly through `opening.codex`. That is very likely
something our stand-in driver does differently, and I will say so once I know.

Nothing else in the PR depends on finding 11.

showell:
Resolution on finding 11, and one thing that survives it.

**The original claim is withdrawn; the cause was ours.** Our harness stands in
for `opening.codex` and skipped the driver's RESOLVE phase entirely --
`build-type-def-map`, `sort-bindings (type-map & checked.all-bindings)`, and
`rewrite-ir-defs` over the IR (opening.codex:823-825). Without the type map the
emitter's `st.type-defs` has no entry for a record declared in the subject, so
`resolve-constructed-ty` fails; without `rewrite-ir-defs` the IR carries
unresolved `ConstructedTy` annotations into emission. The impossible error
counts, the missing diagnostics and a page fault at `__list_snoc+3` all followed
from that one omission.

**What survives is a layout inconsistency in `emit-record`
(X86_64Compound.codex:1751), and I think it is worth closing even though we
reached it by our own bug.** The function picks one of two layouts depending on
whether the record's type resolved. When it does, `build-cce-byte-offsets` ranks
fields by `field-sort-key` -- `width-prefix(field-byte-width) & name` -- and
assigns cumulative widths. When it does not, the `is otherwise` branch ranks by
the raw field name and assigns `rank * 8` via
`emit-store-record-fields-by-list`. Every reader uses the first rule
unconditionally: `cce-byte-offset-and-type` / `accumulate-offset-width-sort`, in
both `emit-field-access` and `emit-record-set-builtin`. The two rules coincide
only when every field is eight bytes wide, since only then does the width prefix
stop affecting the order and the uniform slot match the real width.

`deck-record-repro/probe-record-layout.codex` is the smallest record where they
part, chosen so the narrow field sorts first by name and second by width:

    Box = record { flag : Boolean, items : List Integer }

    resolved:  items @ 0   flag @ 8    (size 9)
    fallback:  flag  @ 0   items @ 8   (size 16)

Built through the fallback, `flag` occupies the slot `items` is read from, so
`b.items` evaluates to the boolean -- 0 or 1 -- and `list-length` of that reads
`[0 - 8]`. That is the fault we spent the afternoon on: `__list_snoc + 3`,
`RDI = 0`, `CR2 = fffffffffffffff8`, called from `bag-add`, whose
`DiagnosticBag` is `List` + `Integer between 0 and 255` + `Boolean`.

**Hedged properly: we could not reach the divergent path through
`opening.codex`.** The probe compiles and prints `3` through the normal driver,
because the type resolves and the first layout is used. We only reached the
second with a driver that skipped RESOLVE, which was our defect and is being
fixed. So this is a latent inconsistency visible by inspection rather than a
demonstrated bug, and it may be unreachable today.

The reason I would still close it: faced with the same unresolved type,
`emit-field-access` records an error and emits `ud2` -- it refuses, and says so.
`emit-record` silently picks a layout no reader shares. Whichever is true of
reachability, one of two small changes seems to settle it: make the `is
otherwise` branch an error like its neighbour, or have it use
`cce-byte-offset-and-type` so the layouts cannot drift apart. The first is
probably right if the branch is genuinely dead, and it converts a future
silent-corruption bug into a diagnostic.

damiant3:
On where the ladder should live: `contrib/zig-ladder/` and `contrib/deck-record-repro/`, a new top-level `contrib/` tree.

Why there and not `codex/plugs/zig/`: the plug tree is gated (`build/build.ps1`, PowerShell-only tooling, no python/sh/zig dependencies), and the ladder is a DDC harness that deliberately runs OUTSIDE that gate with its own toolchain. `contrib/` says exactly that: contributor-maintained, not built by the gate, not held to the fleet's tooling rules, findings flow into the registers (`codex/plugs/plugs-backlog.md`, `codex/compiler/compiler-backlog.md`) the way PR 66's did. Please add a one-paragraph `contrib/README.md` saying so, and re-root the two directories under it on this branch.

The record-layout finding (`emit-record`, X86_64Compound.codex:1751) is going into the compiler register as its own item and will be measured against the seed on our side; if the fallback branch is dead it becomes a refusal like its neighbour, as you suggest.

As with PR 66 the work is re-applied in Perforce and the PR is closed with credit at the push that makes it public, so leave it open.

showell:
Thank you for the clear answer on `contrib/`. We'd like to counter-propose, and we think it gets you everything `contrib/` was reaching for while removing the one thing that makes `contrib/` awkward — a directory inside your tree that your gate is explicitly told to ignore.

**The proposal: the ladder becomes its own repository, `codex-zig-ladder`, maintained entirely on our side. Your repository receives only plug and compiler changes, as small branches cut fresh from master.**

### The split

*Ours, wholly:* `zig-ladder/` (the fourteen rungs, the harness generators, the transports, the sweep), `deck-record-repro/`, every finding write-up, and the READMEs. We maintain it, we document it, we version it, we publicise it.

*Yours, unchanged:* `codex/plugs/zig/ZigEmitter.codex` and anything else under `codex/`. We hold a clone, we patch it there, and we send you PRs — the same traffic you've already taken twice. Six emitter defects so far, all in one file. That is what the relationship has actually consisted of; the ladder was never the payload.

### Why this is better than `contrib/`

**1. Your own reasoning points here.** You said the plug tree is gated and the ladder deliberately runs outside that gate with its own toolchain — python, sh, zig, QEMU, a Linux host. `contrib/` is a way of writing "this is in the repo but not of the repo." A separate repository states the same fact without asking your clone, your diffs, your file counts and your future contributors to carry a tree that nothing in your build will ever touch.

**2. Independence is the instrument's entire value.** The C# plug's own prose says Roslyn is the DDC witness because its lineage is unrelated to yours. The zig arm's claim is the same shape: an independently written code generator agrees, byte for byte. A harness that is vendored into the repository it audits is a weaker witness than one maintained outside it and pinned to a named release. This is not merely tidiness — it is the difference between "the project checks itself" and "an outside party checked the project and published the method."

**3. It converts a hidden assumption into a stated claim.** In-tree, the ladder is "current" by construction and silently so; if a release breaks it, the breakage is yours to discover. Out-of-tree, our master branch names exactly what it is green against, and we hold ourselves to keeping that line true. Today it would read:

> Banked against **Update 45**, seed `270227BE`. 13 of 14 rungs green. `clamp` red — page fault in `__list_snoc`, under investigation.

That is more information than `contrib/` can give you, not less, and it is the sort of statement that is worth something precisely because it is sometimes unflattering.

**4. Your PRs get smaller and more readable.** A depot-bound change is then a handful of files against your current master, with no 120-file workbench alongside it and no long-lived branch history to read past. That is strictly easier for your side to lift into Perforce.

**5. Findings get a durable, citable home.** They currently live in a README on a branch that by design never merges. In the new repository they are first-class and linkable, and the ones that bear on your code still flow into `codex/plugs/plugs-backlog.md` and `codex/compiler/compiler-backlog.md` exactly as you asked — the record-layout finding at `X86_64Compound.codex:1751` being the current example.

### The cross-repo contract — the part you actually need

Your requirement, as we read it, is that your bots can reproduce our results and test our depot-side changes against the right ladder revision. That's a pinning problem, and we'll solve it explicitly:

- **One paragraph in your tree.** We'll still write `contrib/README.md` exactly as you asked, but it points outward: what the ladder is, where it lives, what it is currently banked against, and how findings reach the two registers. One file, no build impact.
- **Every upstream PR we open carries a `Ladder:` line** naming the ladder tag that exercises it, so the pairing is never inferred.
- **Every ladder tag names the upstream commit and seed hash it was banked against**, so the pairing is symmetric.
- **A reproduction recipe on our side, kept working**: clone the ladder at tag *T*, point it at a Codex checkout via a single environment variable, run the sweep, expect the stated rung count. No arguments, no hand-assembly.

That last item is real work on our side — roughly three dozen files currently derive the Codex root by walking a fixed number of directories up, which only works because the ladder sits inside the checkout. Collapsing that to one root-finder is ours to do and is not a cost to you.

### Objections we'd expect, and our answers

*"Split repositories drift."* They drift silently when nothing measures the gap. Here the sweep measures it: a release we haven't re-banked against shows up as a stale line on our master and red rungs when we run it. Drift becomes an observation instead of a surprise.

*"You lose CI coverage of the ladder."* There is none to lose — your gate cannot run python, zig or QEMU today, which is the premise of the `contrib/` exemption.

*"Discoverability."* Solved by the paragraph above, which is the same paragraph you asked for.

### Timing

We expect to have `codex-zig-ladder` standing up and reproducing, with the contract above in place, first thing tomorrow, barring unforeseen problems. Please don't rush anything on your side for it — there are no production users, nobody is paying for this, and nothing here is urgent. If you'd rather have the tree in `contrib/` after all, say so and we'll re-root it on this branch instead. Either way PR 67 stays open per your instruction, and it can sit as long as is convenient for you.

damiant3:
Agreed: `codex-zig-ladder` as its own repository, with the contract as you wrote it (the `Ladder:` line on every upstream PR, the tag naming upstream commit and seed hash, the one-variable reproduction recipe, and the outward-pointing `contrib/README.md` here). Your point 2 is the one that decides it: an outside witness pinned to a named release is worth more than a vendored one, and the stale-line-on-master failure mode is the honest one.

Two things on our side, so nothing hangs on you:

- Once the new repository is up, please cut PR 67 down to `contrib/README.md` alone (or close it and open a fresh one-file PR from master, whichever is less work). That file lands in Perforce like everything else and goes public at the next push, with the credit and the link.
- The record-layout finding is already closed on our side: the unresolved-type fallback in `emit-record` never fired under the gate or the BVT and is now the same refusal `emit-field-access` raises, credited to PR 67 in the change description. It will be public in Update 47.

The `clamp` rung's `__list_snoc` fault on a null list from `bag-add`: if it survives your re-bank on the seed after Update 45, send the smallest subject that reproduces it and we will take it as a compiler item.

No rush on any of it.

showell:
Thank you, and noted on all three. The repository work starts once the current
hunt settles; cutting this PR down to `contrib/README.md` alone is the easier of
the two options and that file is already drafted. Good to hear the `emit-record`
fallback closed as a refusal.

**On the clamp rung: please do not open a compiler item. It is ours, and we
would rather tell you what we found than hand you a false lead.**

`deck-record` in our bundle is compiled as an ordinary function. It is four
bytes in the binary, `48 89 f8 c3`, which is the identity it declares. The gate
at `X86_64Chapter.codex:1147` is computing `deck-record-intrinsic = False` for
our unit, so `emit-deck-record-wrapper` never runs and no call site becomes
`__deck-enter` / body / `__deck-exit`. Everything the compiler deliberately
parks on the deck to survive `emit-all-defs`'s per-function `__heap-restore` is
therefore allocated on the function's own heap and freed at that boundary.
`bag-add` wraps both its cons cell and its record for exactly that reason, so
the second diagnostic a compile records reads a spine that has already been
freed, and `__list_snoc` gets a null list. The prose thirty lines above
`st-add-error` describes this failure precisely, which is how we recognised it.

The cause on our side is a bundler workaround that has outlived its defect: our
scripts rename `deck-record` throughout the unit, which we added when the
intercept still fired on the bare name, before Update 43 replaced that with the
defining-chapter test. We are still isolating which half of the gate fails.
`init-phase-allocator` is now present in the unit, both definitions sit in the
same chapter, and the renamed literal matches the renamed definition, and it
still does not fire, so we are instrumenting rather than guessing. We will
report back either way, and if the fault does survive a clean bundle on the
current seed we will send the smallest reproducing subject as you offered.

Two smaller notes while we are here, neither urgent. First, and offered only as
an observation: the gate fails closed silently. A unit that carries the Phase
Allocator's own `deck-record` but computes the flag as `False` gets a quietly
weaker allocation discipline with no diagnostic, and the symptom arrives much
later as a register dump with no source location. We recognise the flag has to
be able to be `False`, since a plug kernel legitimately ships the identity from
`PlugTypes.codex` and must compile, so this may simply be correct behaviour with
an unlucky consumer. Second, the zig plug has no `poke-byte` row in its builtin
table, though it carries `peek-byte` and `peek-qword`, and `poke-byte` is
reachable from `PhaseAllocator.build` by way of `deck-reservation-guard`. It
fails loudly and names itself, so it is a coverage gap rather than a correctness
one. It looked like the same species as the csharp `peek-16`/`poke-16` row
already open in Update 46.

damiant3:
Update 46 is public at damiant3/NewRepository@adfae029 (seed `12B07296`). Your record-layout finding from this PR is closed in it: `emit-record`'s unresolved-type fallback (the by-list layout) is deleted and the branch is now the same refusal `emit-field-access` raises; the release gate and the BVT never took the branch. Credited to PR 67 in the change and in `docs/PM/Active/GitHubUpdates/GitHubUpdate46.md`.

The PR stays open for the `contrib/README.md` pointer once `codex-zig-ladder` is up, per the plan above. No rush.

damiant3:
Landed: `contrib/README.md` and the `__deck-set` -> `cx_deck_set` fix are in Perforce (main 16627) and will be public with Update 47. The emitter change was re-applied by hand against our current `ZigEmitter.codex` (your branch predates two rounds of zig plug fixes on our side); verified by rebuilding the plug and running two transpiled subjects under zig 0.16.0. `deck-record-repro/` was not taken, since its findings already live in the two registers. Closing this PR now with credit; the ladder repository is the home for the rest, and future emitter fixes as small PRs from master are welcome.

showell:
Cut down to `contrib/README.md` alone, as agreed, and rebased onto Update 46. One file, 17 lines.

**codex-zig-ladder is up:** https://github.com/showell/codex-zig-ladder

It carries its history rather than starting fresh, via `git subtree split`, so the record of how each rung was earned survived the move.

### State it is in

Fourteen of fourteen green against Update 46 (`adfae029`, seed 12B07296), tagged `u46-14of14`. Also fourteen of fourteen against Update 45, and both are banked: `truth/u45/` and `truth/u46/`, one file per rung under the seed that produced it.

**Every rung is byte-identical across the two Updates.** Fourteen subjects, two compilers a release apart, not one byte of emitted image different. Update 46 deleted `emit-record`'s unresolved-type fallback, added CDX3008 and reworked the text builtins, and none of it moved anything the ladder measures. That is the claim the second bank exists to make.

`master` names the seed hash it is banked against, at the top of the README, and refuses to report green against any other.

### What moved where

- The ladder and its READMEs: the new repository.
- The findings record and its probes, twelve of them: moved there too, as `findings/`. Its own opening always said it was discussion material rather than a proposed addition to your tree, so it now has an address instead of a PR. The probes are still Codex chapters and still compile with `build/compile.ps1`. Take any of it in whenever it is useful.
- The `__deck-set` emitter fix: #68, cut from `master` as a small branch with a `plugs-backlog.md` entry (1.38) and a `Ladder:` line naming the tag, the way `contrib/README.md` describes.

### Two things from the Update 46 re-bank, both ours

Neither is a regression in 46. Both were already wrong and already being reported at a severity we read past, and 46 turned them into failures:

- **CDX3008 caught `Core/Collections.codex`** borrowing `TypeBinding` from `Types/TypeEnv.codex` with no cite. Details in #68; the short version is that it is an internal borrow, which `check-subset-cites.ps1` deliberately does not call actionable, and a plug bundle is a subset of that same tree.
- **A duplicate type definition caught us bundling `CCE.codex` twice**, once carried automatically because we cite it and once listed explicitly. Our logs had been printing CDX3006 duplicate-definition warnings about that for months and we treated them as noise. Value duplicates only warn; `CharClass` is a type.

The second one is the more useful lesson on our side. The whole premise of this exercise is that a disagreement is evidence, and we had a standing habit of discarding a category of it.

No rush on any of this.

damiant3:
Update 47 is public at damiant3/NewRepository commit 69cd9ce8 (seed 90646EEB). This is the push that made PR 67's work public: contrib/README.md and the ZigEmitter __deck-set fix landed as Perforce main 16627 and shipped here; the release notes are docs/PM/Active/GitHubUpdates/GitHubUpdate47.md. Thank you, Steve.

PR #68, opened 2026-08-18
zig plug: __deck-set must consume its argument

showell:
`ZigEmitter` maps `__deck-set` to a bare constant and never touches `args`:

```
ZigBuiltinEmitter { name = "__deck-set", emit = \args ctx d ty -> "0" },
```

Answering `0` is right. There is no deck in the zig target, so pointing the deck cell somewhere is genuinely a no-op. Dropping the argument is the defect: the argument is a binding at the call site, and zig refuses to compile a binding whose only consumer disappeared.

```
error: unused local constant
b2: { const deck_base = cx_heap_save(); break :b2 b3: { _ = 0; ...
```

The remedy was already in the same prelude, for the same problem, with a comment that states the rule outright:

```
// address-of answers 0 here by X86_64Compound's own account of
// targets without it; the argument is still evaluated so its binding
// stays used.
fn cx_address_of(v: anytype) i64 { _ = v; return 0; }
```

So this is a second instance of a pattern whose first instance is solved two lines away. `__deck-set` is the only other entry in the builtins table that ignores `args` and takes one. `__deck-enter`, `__deck-exit` and `__deck-pos` also ignore `args` and are all nullary, so the table is clean after this.

### Why it stayed latent

Both callers in the compiler use the address for something else as well, so the binding survives on its second consumer: `build` hands `p` to `deck-reservation-guard`, and `init-phase-allocator` returns `base`. The ladder's `whole` rung transpiles both to zig and compiles clean. It took a caller that only sets.

### What exercises it

`showell/codex-zig-ladder` at tag `u46-14of14`: fourteen rungs green against Update 46 (`adfae029`, seed 12B07296), banked as `truth/u46/` and byte-identical to `truth/u45/` on every rung. Without this fix the ladder's harness prologue does not compile under zig at all, so every rung from `lex` up is evidence for it.

Cut from `master` at Update 46. Two files: the emitter change, and `plugs-backlog.md` 1.38 recording the finding. No changes anywhere else in the tree.

### One note beyond this PR

`build/check-subset-cites.ps1` says only EXTERNAL undefined names are actionable, because internal ones are "the glob assembly model and not a defect". `Core/Collections.codex` is a counterexample. It opens with

```
bsearch-text-pos : List TypeBinding, Text, Integer, Integer -> Integer
```

and cites only `Foreword ListUtils`; `TypeBinding` is in `Types/TypeEnv.codex`. That is an internal borrow, and Update 46's new CDX3008 turns it into a hard error for anyone assembling a subset of the compiler, which is what every plug bundle is. Not a hole in the tool's mechanism, just a gap in what it calls actionable. Carrying the real type in means `TypeEnv`, which borrows `copy-list-with-headroom` from `Unifier` equally uncited, so it cascades. We stubbed the one definition nothing calls; the tidier fix is probably making `bsearch-text-pos` generic or moving it out of Core, but that is a layering call in your tree and not ours to make.


showell:
Closing this: you already landed it. I opened it without reading #67's closing note first, which was there before I started.

`__deck-set` -> `cx_deck_set` is in main 16627 and public with Update 47, re-applied by hand against your current `ZigEmitter.codex`, which is the right way round given the branch here was cut from Update 46 and your main has moved twice since. Nothing to take.

One thing in the body above is not covered by that landing, so recording it here rather than losing it with the close. It needs no action from you and is not a defect claim.

`build/check-subset-cites.ps1` says only EXTERNAL undefined names are actionable, because internal ones are "the glob assembly model and not a defect". `Core/Collections.codex` is a counterexample:

```
bsearch-text-pos : List TypeBinding, Text, Integer, Integer -> Integer
```

cites only `Foreword ListUtils`, and `TypeBinding` is in `Types/TypeEnv.codex`. Internal borrow, so the tool would classify it not actionable. But Update 46's CDX3008 turns it into a hard error for anyone assembling a subset of the compiler, and every plug bundle is one. It surfaced as a build failure the first time the ladder ran against 46.

Carrying the real type in cascades: `TypeEnv` borrows `copy-list-with-headroom` from `Unifier`, equally uncited. We stubbed the one definition nothing in that bundle calls. The tidier fix is probably making `bsearch-text-pos` generic, since it only needs `.name`, or moving it out of Core, but that is a layering call in your tree.

The mechanism of the tool is fine. It is the internal/external split that has an edge, and the edge is exactly where plug bundles live.

Also: `deck-record-repro/` not being taken matches what we did with it. It moved to `findings/` in the ladder repository, where the register can stay ours.

PR #69, opened 2026-08-18
plug bundling: a cite is satisfied by presence, not only when the registry fails

showell:
One change to `Resolve-PlugForewords`, carried in the generator and regenerated.

## The claim

`Resolve-PlugForewords` states the rule at the top of the function:

> A cite is satisfied two ways, and the second one is why the strict pattern needs this: the chapter is RESOLVED through the registry, or it is already PRESENT because `Add-PlugChapter` bundled it.

The code asks the second question only when the first one fails. `$present` is consulted inside the not-found branch, so presence satisfies a cite **only as a fallback**. A chapter that is both present in the unit and resolvable from the registry is added anyway: a second copy, under the quire the cite named, with every definition in it duplicated and a CDX3006 for each.

Hoisting the presence test above the registry lookup makes a cite satisfied by presence, which is what the comment already says a cite is.

## This is prophylaxis, not a bug report

You do not have this today, and I checked rather than assumed. A plug bundle names two chapters (`-Chapters @('ZigEmitter', 'ZigPlug')`) and lets cites do the rest, so the explicit list and the cite closure essentially cannot overlap. You have to hand-list a chapter that something also cites, which is what assembling a *subset* of the compiler ends up doing.

Two of those cost the zig ladder real time. `ListUtils` sat in eleven of its bundles for months at 108-plus CDX3006 warnings a sweep, and we learned to scroll past the class. `CCE` did the same and stayed invisible until Update 46 turned the type-level case into CDX3001 and it became fatal.

## Measured, on seed 12B07296

| what | result |
|---|---|
| all 52 plug bundles built through `Build-TranspilerPlug` | byte-identical |
| the 14 bundles of the zig ladder, plus its other three | byte-identical |
| `arm64` and `riscv`, the two carrying `cites Plug chapter Plug Types` | still resolve, no `exit 3` |
| a bundle that both lists and cites `ListUtils` | 96 duplicated lines gone, nothing else changed |

The fleet numbers come from bundling every plug twice, with the compile step stubbed the way `codex-zig-ladder`'s `cycle.sh` already stubs it. The diff on the reproduction is a pure deletion: 96 lines removed, 0 added, and the 96 are exactly the duplicated chapter.

Worth naming the judgement underneath: two copies under different quires is not meaningless, since resolution is chapter-relative, so a `Parsmi--` chapter mentioning `list-map` gets `Parsmi--ListUtils`'s copy. Someone could in principle want that isolation. CDX3006 warning about it reads as the compiler saying otherwise, but it is your call and not mine.

## On the generated file

`plug-build-lib.ps1` says a hand edit must not be submitted, so the change is in `codex/build/plugbuildlibScript.codex` and the script is regenerated. I compiled the generator with the seed and ran it to confirm the pair:

- before the change, it emits the shipped script line for line, 230 of 230, differing only by trailing newlines that `check-generated-scripts.ps1` trims before comparing
- after the change, the emitted text differs from the shipped script by exactly this hoist and nothing else

The 52-plug measurement above was then re-run against that generated file rather than against a hand edit, so the numbers describe the artifact this PR ships.

## What I have not done

**Run your gate.** It is 46 compiles of PowerShell tooling I deliberately do not run here, and this is a behaviour change for any bundle that would have received two copies, so the right place to find out is your gate rather than my say-so.

**Related, and separable:** `check-generated-scripts.ps1` cannot run on a Linux host at all. `build/compile.ps1` takes its QEMU fallback correctly, but `build/test-compile-batch.ps1` line 98 calls `Start-Process -WindowStyle Hidden` unconditionally and PowerShell on Linux refuses the parameter outright, so the check dies in under three seconds before any compile. That is why the verification above is one `compile.ps1` per generator instead of the batch runner. Happy to send that as its own PR if it is useful.

No backlog entry: the register says an entry is something still missing, and this PR closes what it describes rather than opening it.

The ladder and its findings register live at https://github.com/showell/codex-zig-ladder, tag `u46-14of14`.


damiant3:
Landed in Perforce as main 17251 (red), public with Update 48. Thank you: the hoist is yours as written; the only edit on top is the comment trimmed to two lines, which is house style and nothing else.

What was measured here before it landed, on seed `800A7683` (main 17244):

- The generator emits the shipped script line for line: `check-generated-scripts.ps1 -Only plug-build-lib` reports match, 232 lines, 0 drift.
- All 55 plug bundles in the tree (compile step stubbed, `plug-source.codex` hashed) are byte-identical before and after the hoist.
- The control: a zig bundle that hand-lists `Sort.codex`, which its cite closure also reaches. Old lib 8800 lines with Sort twice; new lib 8742 lines, which is the shipped zig bundle's own count. So the lever moves exactly where you said and nowhere else.
- Internal gate green, `Sut.cdx` byte-identical to the depot seed (not seed-affecting).

On the judgement call you named: we agree with the compiler. Two copies of one chapter under two quires is not isolation anyone asked for, and CDX3006 per definition is the right reading.

On `test-compile-batch.ps1` line 98 (`Start-Process -WindowStyle Hidden` on Linux): yes, send it as its own PR if you have the cycle; it is a real gap and nothing here can see it.

damiant3:
Public now: damiant3/NewRepository commit a061c173 (Perforce main 17251; seed in that commit `800A7683`). The hoist is in `codex/plugs/common/plug-build-lib.ps1` and its generator `codex/build/plugbuildlibScript.codex`. Closing with reference to that commit; the measurements are in the comment above. Thank you.

damiant3:
Closed: landed in Perforce and public at damiant3/NewRepository@a061c173, not merged here because the mirror is downstream of Perforce (docs/Agents/PerforceProcess.md, 'Landing an outside pull request').

PR #71, opened 2026-08-19
zig plug: one arena in the prelude (12.5x memory, byte-identical output)

showell:
One change to `zig-prelude`: every allocation goes through one arena instead of straight to `std.heap.page_allocator`.

## The defect

`page_allocator` rounds every allocation up to a full 4096-byte page (`PageAllocator.map` aligns the length forward). This compiler's average record is about 40 bytes. So:

- a record costs a page
- an x86 instruction, built as a throwaway `List Integer`, costs several (`mov rax,[rbx+8]` is about thirteen)
- one dumped byte costs three, because `cx_show_int` allocates twice and the `& " "` allocates once

`cx_heap`/`cx_hp` looks like an arena and is not one for this purpose: it serves `peek-byte`, `poke-byte`, the `__buf-*` family and the emit workspace. Records, lists, text and closures never touch it.

The result is that transpiled Codex needs roughly 150x the memory per object that bare metal needs. Compiling a 539-line chapter peaked at 3.0 GB and was OOM-killed on a 3 GB machine, doing work the same compiler does inside a 1 GB QEMU guest on the bare-metal side.

## The measurement

Both of the zig ladder's largest units, which transpile the whole compiler and then compile a subject with it:

| unit | before | after |
|---|---|---|
| `fibx` (subjects: fib, then CCE.codex) | 3,045,504 kB, OOM-killed | **242,944 kB** |
| `whole` (subjects: a small chapter, then plug-oracle-arith) | 3,060,352 kB, OOM-killed | **238,080 kB** |

**12.5x**, compile-and-run wall clock 8.4 s.

And the output does not move. Full sweep after the change: **fourteen rungs, both arms, byte-identical to truths banked before it**, including the `pingpong` fixed point. That is the guard on this change — it is output-neutral or it is wrong, and fourteen banked dumps say which.

## What this does not fix, and what should

`cx_heap`/`cx_hp` **is** a faithful bump allocator already: `cx_heap_save`, `cx_heap_advance`, `cx_heap_restore`. It is just not where objects come from, so `cx_heap_restore` rewinds a pointer that nothing important allocates through.

That means `__heap-restore` is a no-op on this plug. It is not a rarely-used builtin: `emit-all-defs` (`X86_64.codex:1082-1089`) brackets **every single definition** in `__heap-save`/`__heap-restore`. On bare metal emission therefore costs the max over definitions; here it costs the sum, and the subject above has 61 definitions. The plug accepts an instruction and discards it.

The honest fix is to unify the two heaps: give the `cx_heap` region an allocator vtable and point everything at it, so `__heap-restore` reclaims what the compiler assumes it reclaims. The trap that makes it more than a rename is worth stating up front — `cx_heap` is an `ArrayListUnmanaged` and reallocating it dangles every pointer already handed out, so the region has to be reserved once and exhaustion must panic rather than grow. Reserving is not committing; lazily faulted pages keep the resident figure where the table above puts it.

Design written up at `findings/zig-heap-unification.md` in https://github.com/showell/codex-zig-ladder. Happy to send that as a follow-up, or to leave it if you would rather take it in a different direction — it changes more of the prelude than this does.

## Not verified here

Your gate. This was measured and swept on the ladder, on Linux, against Update 46's seed (`12B07296`).

Two smaller things noticed while measuring, not touched in this PR: `cx_show_int` allocates twice per number (`allocPrint`, then a second buffer to transcode ASCII to CCE) where one would do, which is about a third of the dump cost on its own; and every x86 instruction being built as a throwaway list is the second-largest term after the dump.


damiant3:
Landed and public: damiant3/NewRepository commit a061c173 (Perforce main 17254, CL 17252 in red's stream). Thank you for this one: the design is yours, the numbers are yours, and they were the reason it went in the same day.

Re-applied by hand against the current `ZigEmitter.codex` since the PR was against Update 46's prelude. What is on top of yours:

- the 22 `std.heap.page_allocator` uses inside `zig-prelude` are `cx_gpa`; `cx_arena_state` / `cx_gpa` are declared right after the `std` import;
- both names are added to `zig-prelude-decls`, so a Codex local that sanitizes to either gets renamed rather than shadowing a file-scope declaration (zig refuses the shadow);
- your comment block is cut to three lines (house style, nothing else).

Measured here on seed `800A7683`: zig plug rebuilt, `plug-oracle-test.ps1 -Only zig` PASS (49 rows now, see issue 72), the emitted subject carries 23 `cx_gpa` and the two `page_allocator` that are the arena's own backing.

Deliberately NOT taken here: unifying `cx_heap`/`cx_hp` with the arena so `__heap-restore` reclaims. You are right that it is a no-op on this plug today and that `emit-all-defs` brackets every definition with it, so emission costs the sum over definitions rather than the max. Your `zig-heap-unification.md` shape (reserve once, vtable, exhaustion panics rather than grows, lazily faulted) reads correct to us; send it as its own PR when the hunt settles and it will be reviewed on its own. Same for `cx_show_int`'s double allocation and the per-instruction throwaway list.

damiant3:
Closed: landed in Perforce and public at damiant3/NewRepository@a061c173, re-applied by hand rather than merged because the mirror is downstream of Perforce.

PR #73, opened 2026-08-19
zig plug: pin a literal-armed switch, and let __linked-list-empty consume its size hint

showell:
Two small gaps, both siblings of code already in ZigEmitter.codex, found by
compiling the plug through itself and still present at a061c173.

**A plain switch with literal arms is unpinned.** An unguarded match whose
arms are bare integer or real literals emits a switch typed `comptime_int`,
which zig refuses to carry into a runtime context. `zig-pin-lit-arms`
already answers exactly this on the guarded chain; this PR applies the same
pin on the plain switch path, one call. In the original formulation of this
fix (measured on the Update 46 seed) the pin fired on about 1% of switches
(7 of 629 in the compiler's check phase, 8 of 805 in text) with zero output
change on every program measured; this PR re-expresses that fix through
your own `zig-pin-lit-arms`.

**`__linked-list-empty` drops its argument.** Any caller whose size-hint
binding has no second consumer then fails zig's unused-local check -- the
`__deck-set` shape again (a binding whose only consumer disappeared). The
`__list-with-capacity` row three lines up already answers this with
`cx_ll_with_capacity`, which consumes and discards the hint and returns an
empty list, so the row reuses it rather than adding a prelude twin.

Verified here: the plug compiles through the a061c173 seed with both
changes, and the hello/recurse/fib warmup oracles are byte-identical to
their bare-metal truths through the rebuilt plug. Our 14-rung ladder is
green on Update 47 with the same two fixes in their original formulation.

### DIFF
--- a/codex/plugs/zig/ZigEmitter.codex
+++ b/codex/plugs/zig/ZigEmitter.codex
@@ -810,7 +810,7 @@ Section: Builtin Emitter
     ZigBuiltinEmitter { name = "__list-with-capacity", emit = \args ctx d ty -> "cx_ll_with_capacity(" & zig-ll-elem ty & ", " & emit-zig-expr (list-at args 0) ctx (d + 1) & ")" },
     ZigBuiltinEmitter { name = "text-compare", emit = \args ctx d ty -> "cx_text_compare(" & emit-zig-expr (list-at args 0) ctx (d + 1) & ", " & emit-zig-expr (list-at args 1) ctx (d + 1) & ")" },
     ZigBuiltinEmitter { name = "char-to-text", emit = \args ctx d ty -> "cx_char_to_text(" & emit-zig-expr (list-at args 0) ctx (d + 1) & ")" },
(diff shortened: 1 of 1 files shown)


damiant3:
Both absorbed, re-applied in Perforce (main 17401) and public at 8f997bd8. Landed exactly as written: `zig-pin-lit-arms` wraps the plain switch path, and `__linked-list-empty` emits `cx_ll_with_capacity(T, hint)`.

Measured here before landing: zig plug rebuilt on seed E45B56F1; `build/plug-oracle-test.ps1 -Only zig` PASS 49/49 against the x86-64 truth; a probe with a plain literal-armed `when` (`is 0 -> 10 / is 1 -> 20 / is otherwise -> 30`) now emits `return @as(i64, switch (n) { 0 => 10, 1 => 20, else => 30, });` and runs under zig 0.16.0 answering 20 and 30. The size-hint row I took on inspection: it is the same shape as the `__list-with-capacity` row three lines up, and `Builtins.codex:187` confirms the Integer argument was always there to consume.

Closing per the usual arrangement (Perforce is primary; nothing merges on GitHub). Rebase onto 8f997bd8 and you have both. Thanks, Steve.

PR #74, opened 2026-08-19
batch compiler: -WindowStyle only on Windows, so the script starts on Linux pwsh

showell:
The gap named on PR 69's close: `Start-Process -WindowStyle` is refused
outright by non-Windows editions of pwsh, so `test-compile-batch.ps1` died
about three seconds in on Linux, before reaching anything real, and took
`check-generated-scripts.ps1` down with it.

`vm-config.ps1` already answers this with a splatted argument table that
adds `WindowStyle` only when `$IsWindows`; this PR applies the same shape at
the batch compiler's one launch site. Changed in the generator
(`codex/build/testcompilebatchScript.codex`) and the emitted script
together.

Verified here: the generator was compiled with the a061c173 seed and run,
and its output matches the shipped script byte-for-byte (280 lines,
trailing newlines trimmed) -- shipped equals emitted, and the diff against
master is exactly the splat. The script now starts on Linux pwsh (empty
list exits 0 in about a second, through vm-config's dot-source) and fails
honestly where codex-vm.exe is absent rather than at parse level. A QEMU
fallback for the batch REPL itself would be its own, larger change and is
not attempted here.


damiant3:
Absorbed, re-applied in Perforce (main 17401) and public at 8f997bd8, generator and emitted script together. `build/check-generated-scripts.ps1 -Only test-compile-batch` reports match, 280 lines, 0 drift, so shipped equals emitted on this side too.

Closing per the usual arrangement (Perforce is primary; nothing merges on GitHub). A QEMU fallback for the batch REPL is, as you say, its own change; if you want to take it, the launch site is now the one place it would go.

PR #75, opened 2026-08-20
zig plug: CCE tiers in the prelude, char-to-text as one byte, and the heap base + deck the corpus oracles observe

showell:
Three commits, a dependency chain, oldest first. The third is the one with a
story; the first two are what it stands on.

## 1. Multi-byte CCE tiers in the prelude

`cce_table` covered codes 0..127 and everything past it collapsed to
`65533`. The prelude now carries the tier-1 and tier-2 block tables:
codes 128..2175 frame as two bytes, 2176..67711 as three, the rest as four
— `cx_cce_frame` writes them, `cx_cce_to_cp`/`cx_cce_to_utf8` read them
back. In the same family: `IrNumLit` now carries its bits through a
`@bitCast` instead of a decimal re-print, `bits-to-real-approx` lands, and
text-to-double mirrors bare metal's own `__text_to_double` digit loop
rather than parsing correctly-rounded — the bits land in `IrNumLit`, so a
parser that rounds better is a parser that diverges from the seed.

## 2. char-to-text is one byte

`char-to-text` emitted a UTF-8 encoding of the char. The bare-metal
contract is one CCE frame appended to the text — with the tiers from the
first commit that is `cx_cce_frame`, and the round-trip
char → text → char is exact for every code the tables name.

## 3. The heap base and a real deck (the corpus oracles observe both)

Two prelude choices were documented as checked-not-assumed, and the depot's
own tests are the check that caught both:

- **The bump pointer booted at 0.** Bare metal boots with
  `mov r10, 6291456`, so a heap address is never 0 there.
  `arith-narrow-proven` asserts exactly that (`__heap-save` proves as a
  structural heap-address fact): expected `mark-ok`, this plug printed
  `mark-zero`. `cx_hp` now boots at bare metal's own base.
- **The deck was a no-op.** Deck position is observable — your
  `deck-bracket-contract` and `deck-record-contract` print it — so a no-op
  deck is a wrong answer, not a simplification. The deck is now the C#
  plug's rule verbatim: the outermost `deck_enter` parks the bump pointer
  in the bivy and swaps the deck pointer in, the outermost exit swaps
  back, and `deck-record` is intercepted by name bracketing its argument's
  evaluation, the way x86-64 and C# both do it.

An aside you may want for the C# plug: its `heap_save` shares the
zero-base defect verbatim (`static long _ptr = 0;`,
CSharpEmitter.codex:637). Its witness path stops at "compiles", so it can
never catch this itself. Source-read only; I have not run it.

## The proof

On Update 47's seed (`90646EEB`): the fourteen-rung ladder, both arms,
14/14 byte-identical with all three commits applied. Then the first full
depot census — all 566 `codex/test` programs through the hosted arm,
diffed against their `.expected`: 162 match, and the three oracles above
(arith-narrow-proven, deck-bracket-contract, deck-record-contract) all
flipped to MATCH with this chain and answer wrong without it.

## The replay

Based on `8f997bd8`. Commits 1 and 2 applied clean; commit 3 had one
conflict, the `zig-prelude-decls` list, resolved as the union of your
arena names and this chain's deck names. The residual diff of the replayed
emitter against my pin is exactly your absorbed content (PR 73's pinned
switch, issue 72's guards) — nothing else moved. Natives built from this
branch by seed `800A7683` spot-verify 5/5: the three oracles above plus
cce-roundtrip and ui-font-cce-order.

## Not verified here

Your gate. The ladder proof ran against Update 47's seed on Linux; the
spot verify against `800A7683`.


damiant3:
Thank you -- this landed in Update 48, public at b643e7cb. Applied verbatim: the file in the release is byte-identical to your branch head (blob 024ee191), as Perforce change 17806, with the attribution in the change description. No tweaks on top.

Verified here before landing, on the release seed: plug rebuilt, plug-oracle-test zig 49/49; arith-narrow-proven through the zig arm now answers mark-ok, and deck-bracket-contract matches its expected byte for byte. Your chain is what flipped both -- and you found them with our own oracles, which nobody here had ever pointed at the hosted arm. That is the part we are taking forward, not just the fix.

Your csharp aside is confirmed (CSharpEmitter.codex:660, static long _ptr = 0) and recorded as plugs-backlog 1.52, credited to this PR.

Release notes: docs/PM/Active/GitHubUpdates/GitHubUpdate48.md, in the same commit. The full release proofs ran against the seed that shipped it (930FF7F1): battery 1541 pass / 0 fail, app sweep clean, poison build clean, and the diverse-double-compile witness holding with zero differing bytes outside the signature region.

PR #76, opened 2026-08-20
zig plug: integer wrap, shadowed-builtin yield, and Char as a CCE code

showell:
Three commits for the zig plug, found by running your test corpus through it and diffing against the `.expected` oracles. All three are fixes on our side; the corpus evidence and the convergence story are below.

**1. Integer arithmetic wraps (`zig plug: integer arithmetic wraps -- Integer is the machine word`)**

Zig's `+`/`-`/`*` are overflow-checked, so the plug panicked where bare metal wraps. The ruling that wrap is the language's rule rests on five converging sources: CodexSubtypes says Integer IS the 64-bit machine word; the IR names behavior wherever it deviates (`IrAddRealTrapping`/`Saturating`) and `IrAddInt` is the plain op; the C# plug emits bare `+` in C#'s default unchecked context; bare metal emits `add`/`imul`/`neg` with no `jo` anywhere; and Foreword's own BloomFilter iterates `hash * 31 + c` unbounded and then tests `if h < 0 then -h` — written against wrap. The commit: `+%`/`-%`/`*%` on the integer rows, `-%` for non-real negation, `cx_ipow` mirroring `__ipow` (negative exponent → 0, wrapping square-and-multiply) in place of `std.math.pow` (panics on both overflow and negative exponents), shift counts masked `& 63` (x86 `cl` semantics; C#'s `<<` on long masks identically).

Corpus effect: the three `panic: integer overflow` subjects (`bloom-spread`, `consistent-hash-balance`, `particle-spread`) now match their `.expected` — and 36 additional programs that previously died in the *hosted compiler itself* (codexir compiled through this plug aborting on checked arithmetic) now compile; 15 of them match outright.

**2. Builtin interception yields to names the module defines (`zig plug: builtin interception yields...`)**

`shadow-builtin-fold` shadows `text-length` and `abs`; the emitted zig contained the user's functions but call sites were hijacked by the builtin-emitter name table (got 3/3/5 where the oracle says 99/99/77). `lookup-zig-builtin` now yields when the unit's own IRDefs know the name, with one paired exemption: a name on the `zig-skip-def` list deliberately has no emitted definition, so its intercept must not yield.

**3. Char is a CCE code (`zig plug: Char is a CCE code, the identity model bare metal and C# share`)**

Follows through on the direction of the CCE-tiers work you absorbed in Update 48: the plug's Char was a Unicode codepoint while `IrCharLit` carried the raw CCE payload, so `char-at s i == 'x'` compared across two alphabets and was false for every letter (`text-fold-indexed`: vowel count 0). The codepoint model is also structurally lossy — CCE aliases canonicalise through any code→codepoint→code detour. This migrates to the identity model bare metal and the C# plug share: `char-code`/`code-to-char` vanish, `char-at` is the same bare read as `char-code-at`, `char-to-text` stores one raw byte, `is-letter`/`is-digit` use the CCE bands your `emit-is-letter-builtin` encodes (13..64 ∪ 97..127, digits 3..12). Codepoints now exist only at the I/O boundary.

Verified by convergence: two probes (`probe-char-literal`, `probe-char-ops`, in the ladder's findings directory) print line-identical output on bare metal and the migrated zig arm, including the accented/Cyrillic second letter band.

**Evidence for the set**: our 14-rung ladder sweep is green over the chain, and the full 566-program census re-bank moved 42 verdicts with zero regressions — the differ, crashed, and hosted-compiler-abort buckets are all empty now (182/566 match; the remainder are our own emitter gaps, not divergences).

**Two source-read asides, not demonstrated, offered in case they're useful**: the Python plug emits `+` on unbounded ints with no 64-bit mask, so it silently diverges from bare metal on any overflow (JS is worse — f64 past 2^53); and `CSharpEmitterExpressions` maps `IrPowInt` to `^`, which is XOR in C#, not exponentiation. `plug-oracle-arith` has no overflow row that would catch either; happy to propose one if you want it.

### DIFF
--- a/codex/plugs/zig/ZigEmitter.codex
+++ b/codex/plugs/zig/ZigEmitter.codex
@@ -855,11 +855,19 @@ Section: Builtin Emitter
 
   lookup-zig-builtin : Text, List IRExpr, ZigCtx, Integer, CodexType -> Text
   lookup-zig-builtin (name) (args) (ctx) (d) (ty) =
-   let len = list-length zig-builtin-emitters
+   if lookup-arity (ctx.arities) name >= 0 & (zig-skip-def name == False) then ""
+   else let len = list-length zig-builtin-emitters
    in let pos = bsearch-zig-builtin-pos zig-builtin-emitters name 0 len
    in if pos >= len then ""
       else let e = list-at zig-builtin-emitters pos
       in if e.name == name then e.emit args ctx d ty else ""
+
+ A name the module itself defines is the module's: the user definition
+ wins over any builtin of the same name, so interception yields before
+ the table is consulted. The arity map is built from the unit's IRDefs,
+ which makes it the module-defined test. The one exemption is paired
+ with zig-skip-def: a definition deliberately not emitted has only its
+ intercept to answer for it, so the intercept must not yield.
 
 Section: Name Mapping
 
@@ -2424,7 +2432,10 @@ Section: Definition Emission
 
  subj-deck-record is the bundler's rename of the compiler's deck-record
  region hint; calls pass through as the argument and the identity def
- itself is skipped rather than emitted as an unusable generic.
+ itself is skipped rather than emitted as an unusable generic. Skipping
+ the def obliges the intercept: lookup-zig-builtin's yield-to-module-
+ definitions rule exempts exactly the names listed here, or every call
+ site would name a definition that was never emitted.
 
   zig-skip-def : Text -> Boolean
   zig-skip-def (n) = n == "subj-deck-record"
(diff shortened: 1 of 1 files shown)


damiant3:
Absorbed into the depot as Perforce main 19045 (all three commits: integer arithmetic wraps with `+%`/`-%`/`*%`, `cx_ipow` in place of `std.math.pow`, shift counts masked `& 63`; the shadowed-builtin yield; `Char` as a CCE code) and published in Update 49, commit bdf0049b on this repository's `master`.

Verification on our side: main carried none of it (the apparent hits were prose); your corpus went 0/5 to 5/5 against the `.expected` oracles with a control run; the zig oracle stays 49/49 either way, which means it cannot catch a `+` put back, recorded in `codex/plugs/plugs-backlog.md` as 1.56 with your C# XOR aside. The release entry is `docs/PM/Active/GitHubUpdates/GitHubUpdate49.md`, "The release head" section.

Thank you, and as before: PRs are absorbed by hand into Perforce rather than merged here, so this closes with the commit that carries it.

PR #78, opened 2026-08-24
compiler-backlog: COMPILER-18, a self tail call whose value is a closure jumps through a poison code pointer

showell:
Ladder: showell/codex-zig-ladder @ finding-38

One backlog row, no code change. Filing a bare-metal fault the ladder found
while building a tail-call tier, since a finding only reaches you this way.

## The fault

`add3 : Integer, Integer, Integer -> Integer` applied to one of its three
arguments gives a closure. Two definitions return that closure and differ
only in how they reach it:

```
make-adder : Integer -> (Integer, Integer -> Integer)
make-adder (n) = add3 n                              -- ((make-adder 5) 20) 22 = 47

count-down : Integer -> (Integer, Integer -> Integer)
count-down (n) = if n == 0 then add3 5
                 else count-down (n - 1)             -- faults
```

The seed-compiled binary prints `applied 47` and then, on the next line:

```
!EXC=0d RIP=a5f000ff53f000ff CR2=a5f000ff53f000ff
        R13=000000000000000a R15=0000000000000019 RBP=000000003ffffff0
```

`a5f000ff53f000ff` is a poison fill rather than an address, so the program
jumped through a code pointer nothing ever wrote. No diagnostic, no CDX
code -- a register dump.

## Why the 2x2 is the point

| return type | reached directly | reached by self tail call |
|---|---|---|
| Integer | fine everywhere | fine (tier 13, five shapes, both arms byte-identical) |
| a function | fine (`make-adder`, 47) | **faults** (`count-down`) |

Neither ingredient is enough on its own, and the working control is in the
same file and the same run as the crash.

## Measured vs read

**Measured:** the table above, on the Update 49 pin `bdf0049b`.

**Read from source, not measured:** `emit-apply`
(`Emit/X86_64Compound.codex:148`) tests `st.tco.active & saved-tail &
is-self-call` first and goes straight to `emit-tail-call`, which shuffles
arguments into the parameter slots and `jmp`s to `loop-top` -- reusing the
caller's frame. That short-circuits ahead of every arity and result-type
decision below it, including the `is FunTy` arm that sends `make-adder`'s
body to `emit-partial-application`. `is-self-call` (`Emit/X86_64.codex:75`)
matches on name alone, with no arity and no return-type test. That names the
asymmetry between the two cells; where `build-partial-app`
(`IR/LambdaLifting.codex:312`) puts a partial application's environment
relative to the frame the `jmp` reuses is what a fix has to read. The ladder
measured the fault, not the emitter, so please treat the mechanism as a
pointer rather than a diagnosis.

## Why nothing here catches it

Every `*-loop` in the compiler returns a value rather than a function, so
the compiler cannot reproduce this on itself.
`codex/test/plug-oracle-arith.codex`'s Lambdas section (added at Update 48
for COMPILER-13) exercises capture, multi-parameter lambdas and application,
but has no self-recursive definition that RETURNS a closure. One row of that
shape there would pin it -- at the cost of a red arm until this is fixed,
which is why this PR does not add it.

## Reproducing

`findings/probe-closure-return.codex` at ladder tag `finding-38` is the
whole thing: six lines of shapes plus a two-line driver. It expects
`applied 47` then `recursed 47`, and stops after the first.

The zig arm cannot corroborate -- it refuses this shape at compile time.
That gap is ours and is tracked on the ladder.


showell:
Closing this in favour of #79, which supersedes it.

Continued measuring after filing this and the framing here turns out to be wrong. The self tail call is not the mechanism -- it is one way to reach a defect in the closure representation itself, and the same corruption is reachable with no recursion at all, sometimes as a silently wrong number rather than a fault. #79 has the corrected claim, two smaller programs, and the mechanism read against current main.

Sorry for the churn. Better a clean replacement than a rewrite of every field in this row.

PR #79, opened 2026-08-24
compiler-backlog: COMPILER-18, a closure carries no remaining-arity, so under-application corrupts silently

showell:
Ladder: showell/codex-zig-ladder @ closure-arity

One backlog row, no code change. Measured on the native x86-64 backend,
seed-compiled, on bare metal under QEMU — no plug anywhere in the path.

## 1. Two spellings of one computation, and the working one is yours

```
  add3 : Integer, Integer, Integer -> Integer
  add3 (x) (y) (z) = x + y + z

    let h = add3 10
    in print-line-uni ("both 42 expected: " & show ((h 20) 12))
    let j = add3 10
    in let g = j 20
    in print-line-uni ("split 42 expected: " & show (g 12))
  end
```

```
both 42 expected: 42
!EXC=0d RIP=e987f000fea5f000 RBX=0000000000600000 R12=0000000000000014
        R13=000000000000000c R14=0000000000600020 R10=0000000000600050
        RDI=000000000000000c RSI=0000000000000014 CR2=e987f000fea5f000
```

The first line is the shape of your fixture at
`codex/plugs/test-input/partial.codex:9-10`, and it is correct. The second
does the same work in two steps and faults. `RDI=0x0c` and `RSI=0x14` are
the 12 and 20 that were supplied; the third argument register was never
written. (`CR2` echoes `RIP` — this is `#GP`, and per
`Emit/X86_64Boot.codex:915` your report shows `CR2` for vector 14 alone, so
that field carries nothing here.)

## 2. A wrong number, with nothing to notice

```
  add3 : Integer, Integer, Integer -> Integer
  add3 (x) (y) (z) = x + y + z

  even-fn : Integer -> (Integer, Integer -> Integer)
  even-fn (n) = if n == 0 then add3 5 else odd-fn (n - 1)

  odd-fn : Integer -> (Integer, Integer -> Integer)
  odd-fn (n) = even-fn n

    print-line-uni ("47 expected: " & show (((even-fn 4) 20) 22))
  end
```

Prints **6291488** — `0x600020`, a heap address, where 47 belongs. No crash,
no CDX code, no register dump.

A crash is a bad afternoon. This is a number that looks like an answer.

## Why this is not an exotic shape

`docs/DevelopersRulebook.md:258-260` states the model: a def "emitted n-ary
is called flat at that arity, under-applied with one arrow per missing
parameter, **over-applied by applying the rest one at a time**."

One-at-a-time is what the language says it does. The native closure is
`[code-ptr][captures]` with no record of how many arguments it still wants,
and a trampoline that only works if it gets all of them at once. The two
cannot both be right.

`codex/plugs/test-input/lambda.codex:19` already drives that path on every
build and passes — because its closure wants exactly one more argument. The
tree tests the boundary and nothing past it. COMPILER-12 records the same
shape on the plug side and warns it can go "right to silently wrong."

## What I could not work out

Program 2 needs the mutual recursion: the flat form `f (n) = add3 5`,
applied the same way, prints 47. I don't know why. Separately, appending a
third statement to an unrelated two-statement program stops its *first*
statement from printing. Both are real, both are in the ladder, and neither
is explained here.

## What I'm asking for

**One ruling, not a patch.** The arity is missing from the object rather
than from any caller, so repairing one call site only moves the boundary.
The choice is whether the closure gains a remaining-arity word so the
trampoline can re-close on under-application — which is what the Rulebook's
model requires — or whether under-application is refused where it is
emitted. That is a representation decision and not mine to make.

**On coverage:** the shape belongs beside `test-input/lambda.codex`, whose
one-remaining case already passes. I have not added it — accepting a red
fixture is your call, not a contributor's.

The row records the citations and what I checked them against. Both programs
are in the ladder at tag `closure-arity` with their recorded output, but they
are printed above so you shouldn't need to go there.


damiant3:
The COMPILER-18 row landed verbatim in commit 111c0fea (CL 19116), and the ruling you asked for is queued to Damian exactly as you framed it: remaining-arity word (the Rulebook's declared model) vs refusing under-application at the emit site. Your source reading verified end to end against the depot (emit-partial-application, the trampoline, both entry paths). When the ruling lands, the fix cites this PR. Account: https://github.com/damiant3/NewRepository/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate50.md. Thank you for routing a representation decision instead of patching a call site.

PR #80, opened 2026-08-24
Plugs backlog: riscv and java do not handle over-application

showell:
One row added to `codex/plugs/plugs-backlog.md` as entry 1.57. Doc-only; nothing else is in the diff.

`docs/DevelopersRulebook.md:256-260` requires a plug that knows the callee's arity to handle three cases — flat at that arity, under-applied with one arrow per missing parameter, over-applied by applying the rest. The rule is unqualified: it binds "a plug", and names the TS/JS family only as plugs that already carry the model. Three plugs implement two of the three.

- **`riscv` has the fix and does not call it.** The named-definition path (`RiscVCodeGen2.codex:583-591`) routes `args < known-arity` to `rv-emit-partial-application`; every other case, `args > known-arity` included, falls into `rv-emit-direct-call` with the whole argument list. Seventy lines below, `rv-emit-closure-over-apply` (`:660-668`) is a correct take/drop over-apply. `grep -rn rv-emit-closure-over-apply codex/plugs/` returns three hits: its signature, its definition, and its own self-recursive tail.
- **`java` never consults arity at all.** `JavaEmitter.codex:158-168` emits a flat call for both the `IrName` root and the `otherwise` root; `lookup-arity` (`:69-70`) has no call site in the file.
- **`arm64` is a near miss, not a defect** — it has `a64-emit-oversaturated-call`, but the arity it consults is a hardcoded builtin-name table, so it does not fire for user definitions.

The zig plug has the same gap. That one is observed end to end — `((even-fn 4) 20) 22` against a one-ary definition emits `even_fn(4, 20, 22)` and zig refuses it at compile time — and it is the ladder's to fix, so it is not in this row.

**What this row does not have.** For riscv and java the runtime consequence is inferred from the dispatch code, not observed. The ladder's host has no PowerShell and no prebuilt plug binaries, so `test-plugs.ps1` cannot run there, and nothing in the harness compiles emitted Java in any case. Per the backlog's own standing hazard about name censuses, that limitation is stated in the row itself rather than left to be discovered.

**Possibly the cheaper half.** `codex/plugs/test-input/partial.codex` covers under-application, saturation and over-application of a *local*, but its only definition is `add3 : Integer, Integer, Integer -> Integer`, which does not return a function — so the branch all three plugs get wrong is unreachable from the corpus. `test-plugs.ps1` judges exit code, non-empty output and text markers without ever compiling what it emitted. One added definition would put all of these in front of a compiler.

**The ask is one ruling:** whether over-application of a named definition is required of every plug that keeps an arity map — in which case riscv wants its dead function wired up and java wants an arity check — or whether some plugs are exempt, and `:258` should say which.

All line citations were re-verified against `5b8091e2`, this PR's base, rather than against the tree they were found on.

Ladder: curried-apply

### DIFF
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -212,12 +212,17 @@ observed end to end: `((even-fn 4) 20) 22` against a one-ary definition
 emits `even_fn(4, 20, 22)` and zig refuses it at compile time with
 `expected 1 argument(s), found 3`. That one is the ladder's to fix and
 is not this row. For riscv and java this entry offers the dispatch code
-and the grep, NOT an observed miscompile: the ladder's host has no
-PowerShell and no prebuilt plug binaries, so `test-plugs.ps1` cannot run
-there, and nothing in the harness compiles emitted Java in any case.
-Per this file's own standing hazard about name censuses, treat the
-runtime consequence as inferred from the emitted shape until someone
-runs a subject through both plugs and reads the output.
+and the grep, NOT an observed miscompile. Per this file's own standing
+hazard about name censuses, treat the runtime consequence as inferred
+from the emitted shape until someone runs a subject through both plugs
+and reads the output. Note what would and would not catch it if someone
+did: `test-plugs.ps1` asserts non-empty text with markers and never
+COMPILES what a plug emitted, so it cannot detect this in `java` however
+often it runs, and by its own prose it does not drive `riscv` or `arm64`
+at all -- the native backends take `-IrInput` and emit the binary wire
+protocol, so they "fail parameter binding and exit 1 in under a second
+having done no work at all" and are deliberately absent from its plug
+list.
 
 **Why none of it was caught, which may be the cheaper half.**
 `codex/plugs/test-input/partial.codex` exercises under-application
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -165,6 +165,77 @@ preceding instruction runs on every path reaching the mov. The guard is in;
 the general case is not. `br` is the standing gap: an indirect branch carries
 no target in its encoding, and this lane emits none today.
 
+**1.57 -- `riscv` and `java` do not handle over-application of a named
+definition, and riscv's correct fix is in the tree with no caller.**
+From the zig-plug ladder (`contrib/README.md`), 2026-08-24.
+`docs/DevelopersRulebook.md:256-260` requires a plug that knows the
+callee's arity to handle three cases -- flat at that arity,
+under-applied with one arrow per missing parameter, over-applied by
+applying the rest. The rule is unqualified: it binds "a plug", and names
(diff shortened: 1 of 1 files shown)


showell:
Correction pushed (`fe0f532d`), before anyone spends time on this.

The row claimed the reporting host has no PowerShell, and offered that as the reason riscv and java were read from source rather than run. **That was false** — pwsh 7.5.4 is installed there and the ladder's own bundler invokes it for every subject. It is not on `PATH`, which is all the check behind that sentence was actually measuring.

**The verdict is unchanged**: riscv and java are still source-level findings and still not observed at runtime. What changed is that the excuse was untrue, so it is removed rather than patched.

What replaces it is a fact about the harness instead of about my machine, and it makes the row's last section stronger:

- `test-plugs.ps1` asserts non-empty text with markers and **never compiles what a plug emitted**, so it cannot detect this in `java` however often it runs.
- It **does not drive `riscv` or `arm64` at all** — by its own prose, the native backends take `-IrInput` and emit the binary wire protocol, so they "fail parameter binding and exit 1 in under a second having done no work at all", and they are deliberately absent from its plug list.

So "nothing checks these two" is a claim about the harness, which is checkable from the tree, rather than a claim about my box, which you had no reason to take on trust.

Actually running a subject through the java plug is a seed compile through our QEMU appliance — a real job rather than a grep. It is queued here; if it produces an observed miscompile I will post it and the row can stop saying "inferred".

showell:
Retracting something I said in the comment above, so it is not left hanging.

I said the java plug run was queued here and that I would post an observed miscompile if one turned up. **That is not coming, and you should not wait for it.** The ladder host has no JDK — `javac` and `java` are both absent — so the emitted Java cannot be compiled here at all. Installing a toolchain to chase this is not a call I should make quietly, particularly given the standing rule this file records about new build environments.

The row is updated (`f1a0…` on the branch) to say so plainly: the finding is filed as a **source-level report**, the verification belongs where the toolchains are, and it names the concrete experiment rather than leaving it as "someone should check" — over-apply a named top-level definition that returns a function, emit Java, and see whether the call site names a method the same file declares with fewer parameters.

Nothing about the verdict changed; `riscv` and `java` were always presented as read from the dispatch code. What changed is that the row no longer reads as "not verified *yet*", which implied a follow-up that was never going to arrive from this side.

The zig instance remains the one observed end to end, and its fix is #83 if it is useful as a worked example of the shape.

damiant3:
The plugs 1.57 row landed verbatim in commit 111c0fea (CL 19117), every citation re-verified against the depot including the three-hit grep. Ruled the same day: the Rulebook's over-application rule binds every arity-keeping plug, so riscv gets its dead helper wired and java gets the arity check, with the observed-miscompile verification your row asks for -- that work is queued in the plugs close-out lane. Account: https://github.com/damiant3/NewRepository/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate50.md. Thank you.

PR #81, opened 2026-08-24
zig plug: a self tail call becomes a loop, not a frame

showell:
**Stacked on #77.** This branch is cut from `zig-plug-heap-unification` (`8cb8a0e4`), so the first 21 commits are #77's and are not this PR's to review. This PR is the last four:

- `6cd40143` the transformation
- `07495229` two zig-shape corrections
- `912daac7` the invariant-parameter rule
- `64d7db8e` an act block is on the tail spine

They touch one file, `codex/plugs/zig/ZigEmitter.codex` (+408/-18). The `X86_64Chapter.codex` change in the combined diff is #77's.

## The defect

Bare metal has tracked tail position since Update 30 (`st-set-tail-pos`), and a self tail call there is a jump. The plug turned every one into a call. Every `*-loop (xs) (i) (acc)` in the compiler has that shape, so depth on this arm grew one frame per element where bare metal's is flat.

`native/zigemit` on the 13.2 MB `ir_to_x86.ir` died in `tokenize_collect`, one frame per token over 3,282,147 tokens: **512 MB died, 2 GiB died, and 3.5 GiB reached only the end of tokenizing.** The same Update 30 commit that gave bare metal tail position gave the python plug its TCO; the zig plug was missed.

## The fix

A definition whose tail positions call itself at full arity emits as `while (true)` with its parameters as loop variables. The spine walks `if`, `let`, unguarded `match`, and (as of `64d7db8e`) the last statement of an `act` block. Any other tail position emits `return <expr>;` unchanged, so an unhandled shape loses the loop and never the meaning.

## Measured

`zigemit` on that same IR completes rc 0 in **27 s at the stock 512 MB stack**, emitting 3,021,734 bytes of zig; 887 of 3,633 definitions emit as loops.

That the program is also *right* is the half a stack fix could have faked. Verified as a whole tree at `64d7db8e`: rebank of 12 units, then **ladder sweep 14/14 green**, diagnostics census unmoved, and all 14 bare-metal truths byte-identical to the Update 49 bank — the emitter change did not reach the image. The bank retaken from that sandbox is a zero-byte diff against the committed one. Both `ir_to_x86` rungs are byte-identical to `truth/u49` through a chain with no QEMU in it.

## Two sub-findings the fix produced

- **The next arguments must go through temporaries.** They read the parameters the loop is about to overwrite, so assigning left to right builds the second argument from the first one's new value. It does not crash — it returns a plausible number.
- **Declining every function-typed parameter was too broad, and cost 10,000 frames.** Zig will not hold a bare function type in a `var`, so the first cut refused any definition having one. `sort-partition` — self tail-recursive in both branches — has a comparator among its six parameters and was left recursing 10,000 deep. A parameter every self call passes back unchanged never varies, so it stays a parameter.

## What `64d7db8e` claims, and what it does not

It is **parity, not a measured stack win.** The python plug has always descended into an act block's last statement; ours treated it as a leaf, so the two plugs disagreed about what a tail position is. Measured honestly: it adds 30 tail loops across 9 of 15 ladder subjects, and on that corpus every one of the 18 functions it reached is the ladder's own dump harness, which ships nowhere. A genuine compiler act-loop of that shape (`print-notices-loop`, `opening.codex`) appears in no ladder subject, so this corpus cannot show an upstream benefit even where one exists. The argument for the commit is that the two plugs now agree; no stack saving is claimed for it.

Ladder: tail-calls


damiant3:
Landed whole in commit 111c0fea (CL 19131), nothing altered. The temporaries-before-assignment rule and the invariant-parameter rule both verified in review; the four-walk agreement hazard you named is now cited in our CL description so nobody drifts them apart. Gate green (plug-binary, plug-smoke, cross-smoke). Account: https://github.com/damiant3/NewRepository/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate50.md. Thank you.

PR #82, opened 2026-08-24
parser: the two top-level scans return their item instead of calling back (COMPILER-19)

showell:
Two commits off `master`: the `Syntax/Parser.codex` restructure, and the `COMPILER-19` row recording it.

## The defect

`scan-top-level` / `try-scan-type-def` / `try-scan-def-header` (the streaming header scan) and `parse-top-level` / `try-top-level-type-def` / `try-top-level-def` (the parse) are each a three-function cycle that goes once around **per top-level definition**, and every edge is a tail call — no frame in either cycle is live when the next is entered.

Nothing in the fleet flattens *mutual* tail calls: `is-self-call` is self-only here, in the python plug, and in the zig plug. So the frames are real everywhere, and invisible on x86-64 only because the kernel stack is generous and the frames are small.

Measured on the compiler's own back-end unit bundled as one document (2,503,544 bytes, 4,511 top-level definitions), through a native binary built by the zig plug: **2,393 nested `scan_top_level` frames** at the shallower cliff, **3,385 nested `parse_top_level`** at the deeper one.

This was only visible because the ladder's `stack_probe.py` censuses the repeated frames on a failing trace instead of reporting pass/fail. The 512 MB stack constant was documented as holding up a lexer prose cycle; what it actually holds up is this scan.

## The change

Source-shaped, and it needs no emitter work anywhere: the two `try-*` functions **return** what they found (a small record — the accumulated lists plus the new state) instead of tail-calling the loop, and the loop tail-calls **itself**, which every existing TCO already flattens.

**Measured with exactly this change applied:** the minimum thread stack the transpiled compiler needs for that document falls from **32 MB to 4 MB**. What remains on the failing trace is `desugar_expr_at` at 297 frames — structural recursion over one expression's nesting depth, not growth in the size of the document.

## Behaviour preserved, at two depths

- The restructured parser compiled that unit to IR, through the plug to a native binary, and both `ir_to_x86` rung outputs are **byte-identical to the Update 49 bank**.
- A full ladder sweep then ran **14/14 green** (1716 s), diagnostics census unmoved, with 13 of the 14 recorded truths byte-identical to that bank. The single mover is the `parse` rung — whose own subject program *is* the edited `Parser.codex`. Its diff is exactly four removed definitions, three added, two added type-defs and 50 line shifts, with **no existing definition changing params, anns or slug**, and both arms agree on the new dump. That is the rung which most directly tests the change.

## On which tree this was measured

The measurement was taken on our zig-plug tail-call branch, because on the Update 49 pin the plug has no self-TCO and the restructure would have flattened nothing there — a null result. Upstream reproduces it through arms it already has, so **the commit here carries the parser change alone**, with none of our emitter work under it.

That rebase is verified rather than assumed: `Parser.codex` at `5b8091e2` is byte-identical to the base the change was measured against, and the file produced by this commit is byte-identical to the verified one.

## One detail a re-implementer must keep

When `parse-type-def` fails, the definition parse re-reads from the **original** state, not from the state the type parser returned. The two functions this replaces did that, and merging them makes it easy to lose.

Ladder: parser-self-tail


damiant3:
Landed in commit 111c0fea (CL 19140) with one tweak: the parse-side prose block was trimmed as a duplicate of the scan-side rationale (house rule: prose about our own code is banned unless load-bearing; the scan-side block and your state-re-read constraint note both stay, the latter verbatim). The restructure itself is untouched and the re-read-from-original-state behavior is preserved and was verified against both old cycles. Full gate green including the text-mode legs your change implicated; measured stack drop is in the CL description with your numbers and your name. Account: https://github.com/damiant3/NewRepository/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate50.md. Thank you.

PR #83, opened 2026-08-24
zig plug: an over-applied definition applies the rest, it does not call flat

showell:
**Stacked on #77.** Cut from `zig-plug-heap-unification` (`8cb8a0e4`), so the first 21 commits are #77's. **This PR is one commit, `835639b7`, touching one file** — `codex/plugs/zig/ZigEmitter.codex`, +92/−2. The `X86_64Chapter.codex` change in the combined diff is #77's.

Sibling of #81, not stacked on it: the two are independent and either can land first.

## The defect

`docs/DevelopersRulebook.md:256-260` requires three cases of a plug that knows the callee's arity — flat at that arity, under-applied with one arrow per missing parameter, **over-applied by applying the rest**. The zig plug implemented two.

Given `((even-fn 4) 20) 22` where `even-fn : Integer -> (Integer, Integer -> Integer)`, it emitted `even_fn(4, 20, 22)` against a definition it had itself emitted as `fn even_fn(n: i64) CxFn2(i64, i64, i64)`, and zig refused: `expected 1 argument(s), found 3`. `emit-zig-apply` looked the arity up, branched correctly on `args < ar`, then let `args > ar` fall into the saturated-call branch — the loop bound in `emit-zig-call-args` is `list-length args`, never `ar`, so `ar` was computed and discarded.

This is #80's rule, on our side of the fence. That row reports riscv and java; this fixes ours.

## Two sites, because the plug held two notions of arity

The second is `zig-closure-make`, reached when a definition is used as a **value** rather than applied. It sizes its wrapper from the **type spine** while the callee is declared at its **parameter count** — `even-fn` has three arrows and one parameter — so every wrapper parameter went flat into a one-ary definition. The emitter already knew these could differ: `zig-untrusted-return-note` prints `"type-val has N arrows for M params"`, and `zig-def-return-trusted` tests `>=` rather than `==`.

That site was predicted from the code and then **measured** rather than assumed. The reproducer needs *two* call sites, because with one the inliner erases the helper and the defect degrades into the ordinary flat call.

## What it emits now

```zig
b5: { const _o5 = even_fn(4); break :b5 _o5.call(_o5.ctx, 20, 22); }
```

Flat at the emitted arity, then the remainder through the returned closure.

**The remainder goes in ONE call, not one argument at a time.** A function type of n arrows is emitted as a single `CxFn<n>` whose `call` takes all n together, so the callee's own return type absorbs the whole remainder by construction. The rulebook's "one at a time" describes a representation this plug does not use — flagging that in case the wording is load-bearing for another backend.

**A generic callee is refused with a marker rather than guessed.** `zig-call-type-args` resolves a definition's type variables from the argument list *and the result type*, and the result type at an over-application belongs to the whole chain rather than the head call, so a variable appearing only in return position would resolve against the wrong type.

## Verified

Sandbox `20260824T220516Z-f40-fix2`, natives built from `835639b7`:

- tier set **GREEN**, 22 tiers
- ladder **sweep 14/14 GREEN** (657 s)
- diagnostics census unmoved (CDX6020 x43)
- the bank taken from that sandbox is a **zero-byte diff against the Update 49 bank** — the change did not reach bare metal, which is what a plug-only change should be able to say

## A detector comes back online

Our tier 14 was excluded from the set while this stood, because one arm would not build. It is back, and it now isolates **#79 / COMPILER-18** with nothing else mixed in: its two controls agree across the arms, and its one disagreeing row is `((even-fn 4) 20) 22` answering 47 here and `not-47` on bare metal — the partial application with no remaining-arity word. That row is admitted in our ledger today, and the day bare metal keeps the arity the arms will agree and the admission will flag itself stale.

Ladder: curried-apply-fix


damiant3:
Landed whole in commit 111c0fea (CL 19133), nothing altered. Both sites verified in review, including the parameter-count-vs-type-spine distinction at zig-closure-make. Your tier-14 detector isolating COMPILER-18 alone is exactly the instrument shape this project keeps asking for. Account: https://github.com/damiant3/NewRepository/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate50.md. Thank you.

PR #85, opened 2026-08-25
zig plug: a self-tail loop reads a top-level definition where the source reads its own parameter

showell:
**Ours, from PR 81, and already in the tree.** Found 2026-08-25 when the ladder's census re-pin moved `dtls-fragment` from `match` to `refused`: `error: unused function parameter`. The refusal is the symptom; the defect under it returns a wrong number with no diagnostic.

`plugs-backlog.md` row **1.58** carries the full account. The short form:

A definition whose parameter shadows a top-level definition gets that parameter renamed to `_arg_<name>`, because zig forbids the shadow. `emit-zig-def`'s loop branch built its body context from `zig-push-tail-renames`, which covers the parameters the loop REASSIGNS and only those — so an invariant slot got no rename at all and fell through to the definition. `dtls-frag-loop` emits calls to `body()` and `msg_type()` where the source reads its own parameters; `dtls_msg_encode` in the same file takes the same two renamed parameters and reads them correctly (`cx_list_len(_arg_body)`). The non-loop emission bound the rename; the loop emission did not.

**Three hunks, and the second is not optional.**

1. The loop branch composes `zig-push-param-renames` under `zig-push-tail-renames`. The tail renames still win for the reassigned parameters because `zig-renamed-scan` reads from the end.
2. `zig-occurs-branches` walks a branch's **guard**. `zig-occurs` drives a discard — `_ = _arg_x;` for a parameter it believes unread — so a read the check can SEE means no discard, an unmentioned `_arg_x`, and zig refusing an unused parameter. That is what `dtls-fragment` did, and it means the obvious shape of this defect cannot return a wrong answer at all. The silent form needs a read the check is blind to, and a guard was one. Without this hunk, hunk 1 turns the wrong answer into a pointless-discard build error.
3. `zig-max-list-len-branches` walks it too. That walk's own prose says it mirrors `zig-occurs`, "same nodes, same reason to visit them, and the same consequence for a node it forgets", and it had the identical hole — shown before fixing: the same 40-element literal emits `@setEvalBranchQuota` in a branch body and none in a branch guard. That one fails loudly at zig compile time and no corpus program reaches it today. Neither walk descends into an effect handler's clauses, which is deliberate and now said out loud in the source: `emit-zig-expr` answers `@compileError` for any `IrHandle` carrying clauses.

**Verified in the order the ladder requires — the tier row first, and it had to fail.** `prim-tailcall` gained a `shadow-guard` row: bare metal 3 against the zig arm's 5, on a loop whose parameter `stop-at` is 3 and whose top-level `stop-at` is 100. Then the fix, then:

- the row green on both arms
- the 22-tier set green: 15 green, 7 noted, 0 unexpected — unchanged in shape from before the fix
- **14 of 14 rungs green** in a sweep, 1589 s
- **`dtls-fragment` back to `match`** in the corpus census, with 320 of the 325 clean programs byte-identical to the previous bank and **exactly one verdict moved**
- every ladder unit emitting byte-identical zig across the two native builds, which is how hunk 3 is shown to change nothing the sweep measures

**Three more programs in `codex/test/` carry the same collision.** `final-batch-test` and `lorawan-encode` had their emitted zig move too. Both stayed `refused` for unrelated reasons (`use of undeclared identifier 'Timestamp'`; a `std.Thread` startFn return type), so neither was ever run and neither could have shown the wrong answer — but the shape is not a peculiarity of `dtls-fragment`.

**Where the instrument was blind, since that is the reusable part.** `prim-tailcall` had exercised loop conversion since the day it was written and was green throughout: no row gave a loop a parameter that shadowed a top-level definition, so the whole class sat outside the tier set. What caught it was the depot's own corpus — a program written by someone with no knowledge of this plug, containing the collision by accident.

Ladder: `shadow-loop-rename`

### DIFF
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -248,6 +248,85 @@ definition is required of every plug that keeps an arity map -- in which
 case riscv wants its dead function wired up and java wants an arity
 check -- or whether some plugs are exempt, and `:258` should say which.
 
+**1.58 -- the zig plug's self-tail loop reads a TOP-LEVEL DEFINITION where
+the source reads its own parameter, and two blind spots had to line up for
+it to be silent. OURS, from PR 81, and already in the tree.** Found
+2026-08-25 when the ladder's census re-pin moved `dtls-fragment` from
+`match` to `refused`: `corpus/dtls-fragment.zig:942:57: error: unused
+function parameter`. The refusal is the symptom; the defect under it
+returns a wrong number with no diagnostic.
+
+**The mechanism.** `dtls-frag-loop`
+(`codex/foreword/encode/DtlsMessage.codex:97`) takes parameters `body` and
+`msg-type`, and the test bundled beside it defines top-level `body` and
+`msg-type` (`codex/test/dtls-fragment.codex:21,23`). Zig forbids a
+parameter shadowing a top-level declaration, so `zig-def-param-name`
+renames them to `_arg_body` and `_arg_msg_type` -- correctly. The emitted
+loop body then calls `body()` and `msg_type()`, which are the top-level
+definitions. `dtls_msg_encode` in the same file takes the same two renamed
+parameters and reads them correctly (`cx_list_len(_arg_body)`): the
+non-loop emission binds the rename and the self-tail-loop emission did
+not.
+
+`emit-zig-def`'s loop branch built its body context from
+`zig-push-tail-renames`, which covers the parameters the loop REASSIGNS
+and only those. An invariant slot is never assigned, gets no loop var, and
+so got no rename at all -- and a parameter of that kind whose name shadows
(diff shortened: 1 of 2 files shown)


damiant3:
Landed: all three hunks verbatim (the guard walks and the rename composition), plugs 1.58 recorded with your verification ladder cited, gate green including the plug binaries. The reusable lesson -- the tier set never handed a loop a shadowing parameter, so the corpus caught what the instrument could not -- is recorded in the row. Closes with the next public update. -- Red, Cobblestone Agent

damiant3:
Absorbed as plugs register 1.58: the self-tail loop reads its own parameter, with your measurement credited in the register row. It ships with Update 50 (commit 8cc80685; report at docs/PM/Active/GitHubUpdates/GitHubUpdate50.md). Thank you!

PR #86, opened 2026-08-25
plugs: a test input for the Rulebook's over-application case, which the corpus cannot currently reach

showell:
**This adds a file that does not run on the x86-64 backend.** That is COMPILER-18, not anything about the file, and it is the first thing to know about this patch. `plugs-backlog.md` **1.59** carries the account.

## The gap

`docs/DevelopersRulebook.md:260` requires three application cases of a plug that knows the callee's arity: flat at that arity, under-applied with one arrow per missing parameter, **over-applied by applying the rest one at a time**. The third case only exists when the definition being over-applied *returns a function*.

`codex/plugs/test-input/` has one input for this area — `partial.codex` — whose only definition is `add3 : Integer, Integer, Integer -> Integer`. It returns an Integer. So every application in that file applies `add3` itself, and the third case is **unreachable from the corpus**. That is the case 1.57 records `riscv` and `java` getting wrong, and the case PR 83 fixed in the zig plug. 1.57 asked for exactly this input: *"One added definition in `partial.codex` would put all of these in front of a compiler."*

## Why a new file rather than `partial.codex`

`partial.codex` runs today; this shape does not. A green fixture quietly turning red is worse than a file that is known-red by construction, so the shape goes in `codex/plugs/test-input/overapply.codex` and `partial.codex` is untouched.

## Measured, both arms, on the exact file this adds

2026-08-25, seed `6CF4A8E0`. The bare-metal arm is the seed compiler under QEMU with **no plug in the path**, so the expected values are not in doubt.

```
bare metal   control-flat: 6
             stepwise: 6291512        <- 0x600038, a heap address where 6 belongs
             <register dump; nothing after that line runs>

zig plug     control-flat: 6
             stepwise: 6
             named-over: 6
             named-over-alt: 7
             after: 15
```

Both of bare metal's failure modes are in one file: `stepwise` is the silent one, `named-over` is the loud one. `compiler-backlog.md` COMPILER-18 gains this as a third and smallest entry point — and it **narrows** that row's open question rather than settling it. `choose` is not recursive and corrupts anyway, so recursion is not the variable; but that row's own control `f (n) = add3 5` is non-recursive with *one* return path and is correct, while `choose` has *two* and is not. The next probe — a single-return-path `choose` — has not been run.

## What this costs you, stated rather than left to be found

- **Not in the standing gate.** `build/build.ps1:814` sets `$smokeSrcs = @('hello', 'record')`, so `plug-smoke` does not read this file. `codex/plugs/test-plugs.ps1` is its only consumer.
- **Its marker leg is vacuous.** The `$markers` table (`test-plugs.ps1:93-98`) has entries for `hello`, `types` and `lambda` only — as is already true for `partial`.
- **UNMEASURED, and this is the real risk:** the harness runs every text plug against every input, and this file puts a definition whose return type is a function in front of roughly thirty emitters that have never seen one from this corpus. Any of them exiting nonzero or emitting nothing turns that cell red. **The reporting host has no toolchain for those plugs and did not run them** — only the zig plug and bare metal, above. COMPILER-13 records four plugs failing the first time a lambda was actually run through them, which is the same class of surprise.

If the file is not wanted before COMPILER-18 is ruled on, it is one `git rm` and the row still stands as a report.

## The finding under the gap

The corpus is not missing this case because nobody wrote the input. Every input that reaches it needs a definition returning a function, and every such input lands on COMPILER-18. **The hole in the plug corpus and the open ruling in the compiler backlog are the same hole from two sides** — an argument for that ruling rather than a separate ask.

Ladder: `closure-overapply`


showell:
**Follow-up measurement, and it corrects this PR.** The row I sent says the open variable is "something about multiple return paths". That is wrong. The variable is **a second call site**, and the branch is irrelevant.

Five programs, one property changed at a time, bare metal each time (2026-08-25, seed `6CF4A8E0`, seed compiler under QEMU, no plug in the path):

| definition | call sites | bare metal |
|---|---|---|
| `sel (n) = add3 1` | 1 | `6` correct |
| `sel (n) = add3 n` | 1 | `12` correct |
| `sel (n) = add3 n`, four statements | 1 | `12, 25` correct |
| `choose (n) = if n == 0 then add3 1 else add3 2` | 1 | `6` correct |
| `sel (n) = add3 n`, four statements | **2** | **register dump** |

The last two rows are the experiment and they differ in exactly one line — whether the third statement calls `sel` again or calls `add3`. That controls for statement count, which COMPILER-18 lists under NOT ESTABLISHED (*"why appending a third statement to an unrelated two-statement program stops its first statement from printing"*).

**Two return paths with one call site is correct.** So it is not the branch, not recursion, and not an argument-dependent capture. With more than one call site `inline-single-caller` cannot fire, so the partial-application object is really built and really entered — and entering it is the defect. Inlined, the closure is materialised at the call site and the trampoline is never reached, which is why every correct row above is correct.

This answers COMPILER-18's own question about why its second measurement needed the mutual recursion: **mutual recursion is one way to defeat the inliner, and a second call site is the cheapest one.** It also means the precondition can be stated in one line rather than left open — *the defect appears exactly when the partial application is not inlined at its call site.*

Worth noting the answer was already written down on the other side of the fence: the ladder's tier-14 prose records that its `control-pick` "survived there only because the pipeline inlines pick and materialises the closure at the call site; even-fn is mutually recursive and nothing can inline it." That was about the zig plug. It holds for the x86-64 backend too.

**What this does not change:** the file this PR adds still does not run on bare metal, for the same reason and with the same measurements. `overapply.codex` calls `choose` three times, so it is a not-inlined case — which is what makes it exercise the shape, and also what makes it red.

**If you want the smallest possible repro for the ruling**, it is two definitions and four statements, and I can send it as a second input or fold it into `overapply.codex` on request:

```
add3 : Integer, Integer, Integer -> Integer
add3 (a) (b) (c) = a + b + c

sel : Integer -> (Integer, Integer -> Integer)
sel (n) = add3 n

  print-line-uni ("control-flat: " & show (add3 1 2 3))
  print-line-uni ("first-call: " & show (sel 7 2 3))
  print-line-uni ("second-call: " & show (sel 20 2 3))
  print-line-uni ("after: " & show (add3 4 5 6))
end
```

Delete the `second-call` line and it answers `6, 12, 15`. Keep it and it dumps registers at `first-call`.

damiant3:
Landed, with a plot twist you could not have seen: both of your failing lines were fixed at our head HOURS before your PR arrived (COMPILER-18 and COMPILER-20 closed together in one CL), so your known-red file measured GREEN on bare metal here -- all five lines matching your zig arm exactly. Your 'multiple return paths, not recursion' narrowing is the same conclusion the internal closure reached the same day; the register row now records both derivations. The file lands as the green corpus fixture for the Rulebook's third case, your test-plugs cost notes recorded verbatim, the thirty-emitter sweep still honestly unmeasured. Closes with the next public update. -- Red, Cobblestone Agent

damiant3:
Absorbed as plugs register 1.59, and your fixture ships in this push as codex/plugs/test-input/overapply.codex - measured red at your seed and green at head, exactly as designed. It ships with Update 50 (commit 8cc80685; report at docs/PM/Active/GitHubUpdates/GitHubUpdate50.md). Thank you!

PR #88, opened 2026-08-25
plugs 1.61: no run.ps1 consults the VM host selection, so no plug runs on Linux

showell:
Doc-only: one entry in `codex/plugs/plugs-backlog.md`. No code changes, and deliberately no patch — see the last paragraph.

`build/vm-config.ps1:14-16` states that "QEMU is the fallback and the only host on Linux/WSL. Both paths are live", and implements the choice at `:21`. **None of the 56 `codex/plugs/*/run.ps1` reads it**: 38 delegate to `build/plug-run.ps1`, which hardcodes `tools\codex-vm.exe` at `:49` with no fallback; 8 hardcode the same path themselves; 10 take `$script:CodexVmBin` from the sourced config and skip `$script:UseCodexVm`. On Linux none of them can start a VM.

The failure also names the wrong thing. `Start-Process` on a missing path is terminating under `$ErrorActionPreference = 'Stop'`, so control leaves the `try` for the `finally` at `:167`, where `:168` reads an unset `$proc` under `Set-StrictMode -Version Latest` — and that exception replaces the original. What a user sees is `The variable '$proc' cannot be retrieved because it has not been set`. Reproduced in isolation with those four lines.

`build/compile.ps1` shows the intended shape: `:218` is `if ($script:UseCodexVm)`, `:239` falls back to `Invoke-VmCompileFallback`. Measured on the same Linux box, that leg works — a `.codex` source to IR-CCE in four seconds in a `qemu-system-x86_64 ... -m 3072` guest.

The transport is the part that is not a copy-paste: under codex-vm the guest dials the host's TCP listener, while the QEMU consumers in `vm-config.ps1` read the serial wire (`:835`). The entry carries the ladder's working recipe instead — it runs this transport daily against the zig plug.

**The ask is one ruling: is Linux a supported host for RUNNING plugs, or only for building them?** Either answer is a small change; they are different changes.

Hedged where it matters: we have not tried a fix on Windows and cannot, and `build/plug-run.ps1` is generated, so any change belongs in `codex/build/plugrunScript.codex`.

Ladder: `plug-run-no-vm-host` (https://github.com/showell/codex-zig-ladder)

### DIFF
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -248,6 +248,109 @@ definition is required of every plug that keeps an arity map -- in which
 case riscv wants its dead function wired up and java wants an arity
 check -- or whether some plugs are exempt, and `:258` should say which.
 
+**1.61 -- no `run.ps1` consults the VM host selection in the config it
+sources, so no plug can be run on Linux.**
+From the zig-plug ladder (`contrib/README.md`; the repository is
+https://github.com/showell/codex-zig-ladder), 2026-08-25, measured on a
+Linux host against the tree at git `0c4327d5`, the Update 50 interim
+push. Not a plug defect -- it is how every plug reaches a VM.
+(1.58-1.60 are claimed by ladder branches in flight.)
+
+**The contract.** `build/vm-config.ps1:14-16`: "codex-vm (the WHP
+hypervisor) is the primary and is Windows-only; QEMU is the fallback and
+the only host on Linux/WSL. Both paths are live: a missing codex-vm is
+not an error, and the hard failure is reserved for having NEITHER host."
+The file implements exactly that -- `:21` chooses (`$script:UseCodexVm`),
+`:22-52` discovers a QEMU, `:55-56` is the error for having neither.
+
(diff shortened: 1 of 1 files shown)


damiant3:
Landed as register row 1.73 (renumbered for the same same-day collision as your others) -- and with a ruling attached: Damian read your row and ruled the fallback SUPPORTED on every host, Windows included, not just Linux. The fix is routed to the lane that owns the runner scripts, leverage-first (the shared plug-run.ps1 covers 38 of the 56 in one edit). Your instinct to send the measurement without the patch was exactly right; the fan-out was ours to choose. Closes with the next public update. -- Red, Cobblestone Agent

damiant3:
Absorbed as plugs register 1.73, and it drew a ruling: Linux/QEMU is a supported plug host - plug-run honors the VM host selection and plug-smoke gained a QEMU arm with a byte-identical cross-host requirement. It ships with Update 50 (commit 8cc80685; report at docs/PM/Active/GitHubUpdates/GitHubUpdate50.md). Thank you!

PR #89, opened 2026-08-25
plugs/common: IRTextParser takes the ir- prefix its counterpart already uses

showell:
The two ends of the IR text wire are written to opposite conventions:

    IRTextEmitter.codex     106 defs,  99 prefixed 'ir-'
    IRTextParser.codex      113 defs,   0 prefixed 'ir-'

So the parser owns the plainest names in the tree -- tokenize, parse-expr,
parse-type -- which are exactly the names a compiler's own lexer and parser
own. Nine of its definitions collide with codex/compiler/Syntax:

    parse-expr                ParserExpressions.codex:4     IRTextParser:686
    parse-act-stmts           ParserExpressions.codex       IRTextParser
    parse-handle-clauses      ParserExpressions.codex       IRTextParser
    parse-type                Parser.codex                  IRTextParser
    parse-type-atom           Parser.codex:249              IRTextParser:262
    parse-type-args           Parser.codex:431              IRTextParser:352
    parse-record-fields-loop  Parser.codex:785              IRTextParser:374
    scan-string-body          Lexer.codex:259               IRTextParser:90
    tokenize                  Lexer.codex                   IRTextParser

They are pure name coincidences -- two parsers for two different languages,
sharing no type (ParseState/Token/LexState against SExp/Text) -- but the
consequence is real: no program can carry the compiler and the IR text
parser at once. We hit it building a single-binary Codex-to-zig transpiler
that wants both.

This renames those nine and their -suffix families, 31 definitions, with the
ir- prefix. Pure rename, no semantics: 156 insertions and 156 deletions.

parse-ir-chapter is deliberately NOT renamed. It is the public entry point
and 54 plug bodies call it; it is also already unambiguous.

Three call sites outside the chapter move with it:
  CSharpPlug.codex:108              tokenize    -> ir-tokenize
  CSharpEmitterExpressions.codex:184,192  parse-type -> ir-parse-type

Not touched: codex/test/apps/annotation-reader-test.codex:9 contains the
string literal make-target "Parser" "parse-expr", which names the COMPILER's
function and is not a call into this chapter.

VERIFIED ON THE ZIG ARM, and only there. The ring plug, zigemit, codexir and
our combined binary were all rebuilt from this source; 85 programs
(codex/plugs/test-input, the ladder's warmups, its prim-* and probe-*
subjects) transpile byte-identically through codexir | zigemit and through
the single binary; and that binary now emits its own 2.8 MB bundle as
2,273,737 bytes of zig identical to what the seed-plus-ring-plug path emits
from the same source.

WE CANNOT RUN THE C# ARM -- this host has no toolchain for it, and we are
not claiming otherwise. The change there is three call sites and no
semantics, but it is unrun. Ladder: codexzig-fixed-point

### DIFF
--- a/codex/plugs/common/IRTextParser.codex
+++ b/codex/plugs/common/IRTextParser.codex
@@ -1,6 +1,6 @@
 Chapter: IR Text Parser
 
-   1. tokenize : Text -> List SExpToken
+   1. ir-tokenize : Text -> List SExpToken
    2. build-tree : List SExpToken -> SExp
    3. walk-* : SExp -> {IRChapter, IRDef, IRExpr, ...}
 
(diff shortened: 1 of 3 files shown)


showell:
**Test tree, for the record.** These numbers were measured against our fork's stack rather than bare `upstream/master` — `0c4327d5` plus PR #85 (the finding-42 emitter fix) and this branch. That is the optimistic version of what we hope goes into Update 51; measuring against bare upstream would measure a compiler already missing our own fixes.

On that tree the zig arm is now checked end to end against the depot's own oracle: all **181** well-behaved corpus programs (`clean` + `match` in our census — the plug refuses no construct and the pipeline's output already agreed with `.expected`) build, run, and match their `.expected` files, and all 181 emitted `.zig` files are byte-identical between the two-process pipeline and our single-binary transpiler.

Specifically for this rename: rebuilding the natives with it stashed emits **byte-identical** zig, so it moves no output. The C# call sites remain unrun here, as the description says.

damiant3:
Landed at head as a clean patch apply, gate green including the plug binaries and cross-smoke. The csharp lane owner confirmed nothing in flight against the renamed identifiers. Closes with the next public update. -- Red, Cobblestone Agent

damiant3:
Absorbed at head: IRTextParser carries the ir- prefix its counterpart uses. It ships with Update 50 (commit 8cc80685; report at docs/PM/Active/GitHubUpdates/GitHubUpdate50.md). Thank you!

PR #90, opened 2026-08-25
COMPILER-20: implicit record type parameters are derived by the serialiser, not carried on the AST

showell:
Measured on the zig-plug ladder, 2026-08-25. IRTextEmitter derives a
record's implicit type parameters at write time; the ATypeDef the front end
holds has tparams = [] for a type like foreword/core/Sort.codex's
SortPartition, which declares no parameter list and leaves 'a' free in its
fields. Every plug reads the text and sees a complete type; anything reading
the AST does not, and emits code that does not compile.

Doc-only: one row in codex/compiler/compiler-backlog.md, no code change --
the fix touches the type checker and we can verify only the zig arm.

Ladder: codexzig-fixed-point


showell:
**Test tree, for the record.** These numbers were measured against our fork's stack rather than bare `upstream/master` — `0c4327d5` plus PR #85 (the finding-42 emitter fix) and this branch. That is the optimistic version of what we hope goes into Update 51; measuring against bare upstream would measure a compiler already missing our own fixes.

On that tree the zig arm is now checked end to end against the depot's own oracle: all **181** well-behaved corpus programs (`clean` + `match` in our census — the plug refuses no construct and the pipeline's output already agreed with `.expected`) build, run, and match their `.expected` files, and all 181 emitted `.zig` files are byte-identical between the two-process pipeline and our single-binary transpiler.

Specifically for this rename: rebuilding the natives with it stashed emits **byte-identical** zig, so it moves no output. The C# call sites remain unrun here, as the description says.

damiant3:
Absorbed into the internal register as COMPILER-26 (renumbered: the internal register minted 20-25 independently the same day you filed). Both citations re-verified against head before landing. Ships with the next public update, and this PR closes with it. Thank you -- the write-time-derivation catch is exactly the class of thing a second emitter consumer was always going to hit first. -- Red, Cobblestone Agent

damiant3:
Recorded in the compiler register as COMPILER-26 (our numbering): a record's implicit type parameters are derived by the serialiser rather than carried on the AST. It ships with Update 50 (commit 8cc80685; report at docs/PM/Active/GitHubUpdates/GitHubUpdate50.md). Thank you!

PR #91, opened 2026-08-25
COMPILER-21: the deck reservation is advisory and overrunning it faults at 200%

showell:
__heap-advance reserves; nothing enforces the reservation. The tracer
prints negative headroom on every step past the boundary and no path acts
on it, so the program allocates to twice the reservation and then takes a
General protection exception in cx_list_at.

Reproducible in ten seconds with no VM by lowering the literal in an
emitted hosted tool and rebuilding. The cursors-meeting overrun IS guarded
and refuses by name, which is why this looks like a gap rather than a
policy.

Doc-only: one row in codex/compiler/compiler-backlog.md.

Ladder: deck-reservation-advisory


damiant3:
Absorbed into the internal register as COMPILER-27 (renumbered from your 21 for the same collision). Your 421/512 proximity measurement moved this from latent to soon; the bare-metal analog of the boundary is now the row's first question. Ships with the next public update, and this PR closes with it. -- Red, Cobblestone Agent

damiant3:
Recorded in the compiler register as COMPILER-27 (our numbering): the deck reservation is advisory, with the 421/512 MB proximity measured and registered. It ships with Update 50 (commit 8cc80685; report at docs/PM/Active/GitHubUpdates/GitHubUpdate50.md). Thank you!

PR #92, opened 2026-08-26
zig plug: six emitter defects fixed, two reported open; corpus refusals 112 -> 24

showell:
Six defects in the zig plug, found and fixed on 2026-08-26, plus two
reported open. Cut from Update 50 with **no stack under it** — all eight
prior PRs landed in that release, so every measurement below is against
the bare release rather than against a patched tree.

**Headline, from `corpus_run.py --run`, which builds and runs each
program and diffs against its `.expected` rather than counting markers:**

```
corpus match      183  ->  269
corpus refused    112  ->   24
ast/allcycles.sh  14/14 on every chain, final sweep 23:05
```

Four verification chains, six legs each (natives, a type-variable case
matrix, the corpus, the `codexzig` self-emission fixed point, eleven
ported Roc snippets, and the bare-metal sweep). Green throughout.

## What is in it

| row | defect | reach |
|---|---|---|
| 1.85 | the type-variable scope test this row asked for, and its measurement | markers 40 → 0 |
| 1.86 | a refusal strands the parameters it consumed; zig reports the stranding, not the refusal | 4 programs |
| 1.87 | `show` dispatches five ways on the argument's type; the plug implemented one arm for all five | 42 of 113 refusals |
| 1.88 | emitted `main` spawns `opening` directly and zig refuses a thread entry returning a value | 40 programs |
| 1.89 | a unit family was mapped to `void`, erasing the payload | 6 programs, 11 more open |
| 1.90 | the prelude's own locals and parameters shadow user top-level names | 66 reserved names — **open** |

Also included: eleven test programs ported from Roc's evaluator suite,
carrying Roc's own expected values. They are oracles written by people
who have never seen this emitter, and two of the six defects above were
found by them rather than by the corpus — including 1.87, which had been
sitting in our own corpus output as its largest single refusal class for
as long as the corpus has existed.

## Two things you should know before reading the diff

**1.87's Boolean fix duplicates work you have already done.** You told us
on 2026-08-26 that you had fixed the Boolean literal-pattern class in
three plugs including your copy of this one. `upstream/master` was still
`8cc80685` when this branch was cut, so the commit is here. Ours guards
`TextTy` — a Text literal pattern spelled `"True"` is left alone — which
yours may or may not. Take whichever you prefer; we would rather flag the
overlap than have you find it in a conflict.

**1.90 ships a change that fixes nothing, on purpose.** It renames four
prelude locals. The two programs it targets still refuse, because the
rename moved the error to a function parameter of the same name. It is
included because it is exactly what was measured, and because measuring
it is how we learned the class is 66 names rather than the 45 we first
counted — our first extraction searched `const` and `var` and never
looked at parameters. The row costs both candidate real fixes and takes
neither.

## What is deliberately NOT in it

Three changes written the same day and left out for cause:

- **A lambda type-recovery rule.** Written, never built, and it carries a
  known gap recorded in its own prose: it guards `let` and nested-lambda
  rebinding but not match-branch binders, so a binder sharing a
  parameter's name would be read as evidence about the parameter. That
  produces a wrong recovered type rather than a refusal, which is the bad
  direction.
- **The other half of 1.89** — declaring unit families so
  `emit-zig-atype`'s verbatim name becomes correct. Written an hour
  before this branch, never built.
- **The 1.90 class fix.** Not written.

## Method note

Every fix here carries a prediction written into the finding *before* the
build that tested it. Two of five were wrong, and both wrong ones are
recorded in the rows rather than quietly dropped — 1.88's predicted a
wave of newly-visible failures behind the 40 it unblocked and got two,
and 1.90's predicted two programs would start passing and they did not,
which is how the 66-name surface surfaced at all.

Ladder: `u50-emitter-batch`

---
*Written by Claude (Claude Code), working with Steve Howell on the zig-plug ladder. Posted from his account. The measurements are the ladder's; the prose and the judgement calls about what to include are mine, and Steve reviewed them before this went up.*


showell:
## Corpus headline re-scored: `match 269 -> 267`, and 13 programs newly excluded

You flagged these counts as provisional pending our error-gate work. That work is done and here is the number.

**What was wrong.** `ast/CodexIrHarness.codex` called `check-chapter`, bound `cr.state`, and never consulted the diagnostic bag, so `native/codexir` emitted IR for a program with compiler errors and said nothing. `corpus_run.py` runs that tool, so those programs were transpiled, built, run and scored anyway.

**The gate**, taken from one shared place rather than copied into both generators (`BAG_MERGE` / `halt_gate()` / `halt_formatter()` in `ast/emit_harness.py`). The already-gated zig harness regenerates byte-identical after the refactor, so its half is provably inert. `check_harness_gates.py` now compares both generated harnesses to each other **and** to `opening.codex`, because agreeing with each other is necessary and not sufficient — our `ir-emit-roots` list once drifted in both at the same time, and being wrong together looks exactly like being right.

**Re-scored, same tree (`cab52a35`), gate as the only delta:**

```
before   606 programs: clean 326, markers 264, unresolved 16
         match 269, refused 24, no-expected 30, hardware-only 2, crashed 1
after    606 programs: clean 317, markers 260, unresolved 16, zigemit 13
         match 267, refused 24, no-expected 23, hardware-only 2, crashed 1
```

**13 programs carry compiler errors.** Nine left the clean set:

| program | previous verdict |
|---|---|
| `quotes-gate` | **match** |
| `quotes-parse` | **match** |
| `class-op-no-instance` | no-expected |
| `effect-launder-fork` | no-expected |
| `effect-launder-map` | no-expected |
| `effect-launder-record` | no-expected |
| `let-effectful-bug` | no-expected |
| `mutable-alias` | no-expected |
| `parser-resync` | no-expected |

So **exactly two verdicts were contaminated**. The other seven asserted nothing, having no `.expected`. No program that stayed in the clean set changed verdict — the diff is 0 movers plus 9 removals.

**Corrected headline for this PR: `match 183 -> 267, refused 112 -> 24`.**

**What is not affected.** The six emitter findings' reach numbers are counted by error signature, not by the clean set — 42 of 113 for `show`, 40 for the thread entry, 6 for units, 4 for the stranded parameters, 2 for the Boolean patterns. Those are unchanged. The rows in `plugs-backlog.md` quote per-defect counts rather than the corpus headline, so they stand as written.

**An independent check that came free.** Our other harness's gate finds 13 halts on this corpus after a `cite_resolve` fix earlier the same day took it from 41 to 13. This new gate, in a different harness, independently finds 13. Two gates, same number — which is precisely the agreement that was missing when the ungated harness was telling us things about your type checker.

Thanks for calling the counts provisional rather than taking them; you were right to, and it is a better habit than the one we brought to this thread.



---
*Written by Claude (Claude Code), working with Steve Howell on the zig-plug ladder. Posted from his account.*

damiant3:
Reviewed whole and absorbed into Perforce (CL 19961, main 19963). The design and every measurement are yours; verdicts below, per your ask for issues on the PR.

**What we verified:**
- All six fixes read against source and matched your register rows exactly, including the honest partials (1.85's scope test, 1.89's half, 1.90 open).
- **All twelve `.expected` files re-measured on bare metal** against the shipped seed `C3181693` before being trusted as cross-backend truth: **12 of 12 match**. They are now depot tests and gate with everything else.
- Gate green (192.1 s standing gate).

**Tweaks made on top, with reasons:**
- Your `zig-lit-pat-text` replaced our own fix for the same defect (our compiler lane landed `zig-bool-lit-text` yesterday, independently, after your email lead). Yours won the merge because it dispatches on the pattern's TYPE first, so a genuine Text literal pattern passes through where our unconditional spelling swap would have corrupted it. Same finding, your better fix.
- `TechnicalDetails.md` test-file count re-measured 1,676 -> 1,688 for your twelve pairs.

**Noted, no action needed from you:**
- The corpus headline (match 183 -> 269, refused 112 -> 24) we treat as provisional pending your harness error-gate work, per the email thread -- not a mark against this PR.
- Your 1.88 note about our C# plug discarding an effectful `opening`'s value is recorded in our plugs register as a lead, with your credit.
- `zig-subst-arg-type` being uncalled is now ours to decide; you were right not to delete it.

This PR stays open until the next mirror push makes the work public, then closes with that commit per our usual procedure. Thanks for another excellent batch -- the refusal-over-guess discipline throughout (reals, entry types) is exactly the house style.

showell:
Thanks — and one pointer, since our comments crossed: the harness error-gate work is **done and measured**, and the re-scored number is [in the comment just above](https://github.com/damiant3/Cobblestone/pull/92#issuecomment-5432571472). Short form so you needn't chase it:

**13 corpus programs carry compiler errors. Two verdicts were contaminated** (`quotes-gate`, `quotes-parse`); the other seven had no `.expected` and asserted nothing. Corrected headline: **`match 183 -> 267, refused 112 -> 24`.** Zero movers among the programs that stayed — the whole diff is nine removals from the clean set. So the counts are no longer provisional, and 267 is the number that survives.

Two notes on your verdicts.

**On the twelve `.expected` files:** 12 of 12 on bare metal is more reassuring to us than to you. Those values came from Roc's suite, adapted, and every run *we* made compared them against our own arm — we had queued the bare-metal check as work we owed and you got there first. Worth stating plainly that they were unverified cross-backend truth until your measurement, not after it.

**On `zig-lit-pat-text` winning the merge:** that type guard exists only because a cold read demanded it. The first version swapped the spelling unconditionally, and the reviewer flagged the `TextTy` case as a trap for later — both spellings refuse loudly today, so nothing was *currently* wrong. It went in on that argument alone, with no failing test behind it. Twenty-four hours later it was the difference between two fixes. We'd rather record that than take credit for foresight we didn't have.

---
*Written by Claude (Claude Code), working with Steve Howell on the zig-plug ladder. Posted from his account.*

damiant3:
This work is public as of tonight's mirror push: commit a0425e10, on top of the Update 51 release earlier this evening. The absorption details and review verdict are in the comment above; the release-notes entry is the Addendum section of docs/PM/Active/GitHubUpdates/GitHubUpdate51.md in that commit, which credits this PR by number.

Six fixes, two honestly-reported open rows, and twelve new cross-backend truths in the depot's own test tree -- all landed with your measurements intact. Closing with thanks; findings 1.85's residue, 1.89's second half, and 1.90's prelude-shadow class are recorded in the plugs register as open work with your name on the finds.

damiant3:
You were right, and it is now actually public: commit 012a9d2e carries the emitter (zig-lit-pat-text present, 3315 lines against the 3002 you diffed at a0425e10).

What happened: the absorb landed in our depot in full on 8/26, but the copy-up to the mirror branch dropped ZigEmitter.codex (and the register rows) to a missed "Can't clobber writable file" refusal, so the push shipped the claim without the content. Your second observation was just as correct: our 12-of-12 re-measure was x86-only and could not have observed the emitter either way. Before the repair landed we graded the zig arm with a plug rebuilt from this emitter: 4 of 12 pairs run and match (roc-early-return-predicate True, roc-recursive-var 6, roc-returned-closure 9, tvar-in-declared-type 73), and 8 refuse with the designed clean markers, matching your rows exactly.

The full account is in docs/PM/Active/GitHubUpdates/GitHubUpdate51.md (addendum correction), and docs/Agents/PerforceProcess.md's P-CLOBBER row now names the copy-up variant plus a file-list accounting check so the next one fails loudly instead of shipping. Thanks for reading the mirror closely enough to catch us publishing a claim -- both halves of your report were exact. 1.89's other half and 1.90 stay open in the register.

PR #93, opened 2026-08-27
The compiler computes lambda parameter types and the IR does not carry them

showell:
*Written by Claude, on Steve Howell's account and at his direction. Found by
the zig plug's ladder; `Ladder: h2-lambda-span`.*

Three commits: the four-line side channel, the CST span that makes it work,
and the `COMPILER-30` row in `codex/compiler/compiler-backlog.md`.

---

## The claim

The IR discards type information the compiler has already computed, and the
discard is **invisible to almost every backend in the fleet** — so nothing in
the project's own verification can see it. Bare metal does not care whether a
value is an `Integer`, a `Text` or a pointer; it is a register either way.
The two artifacts this project trusts most — the x86 reference and the C#
diverse-double-compiling witness — are, by construction, the two that cannot
feel this loss. The result is an IR shaped around backends that erase types,
validated by backends that erase types, and **it has already produced wrong
code in backends that do not.**

We are not asking for the zig plug to be given special treatment. We are
reporting that zig is the first backend in the fleet that can neither erase
the missing type nor guess it, which is why it is the one surfacing this.

## The specific fact

`infer-lambda` (`Types/TypeCheckerInference.codex:488-498`) computes the
lambda's complete function type — `wrap-fun-type (pr.param-types) lam-row
(br.inferred-type)` — and returns it as `inferred-type`.
`bind-lambda-params` (`:506-516`) binds each parameter to a fresh type
variable, unification solves them, and the solution is sitting in the
`UnificationState` that call returns.

Lowering does not ask. `lower-lambda` (`IR/Lowering.codex:721-728`) peels the
parameter types off the type its **context expects** (`:724`), and
`peel-fun-param` (`Types/CodexTypeHelpers.codex:4-10`) answers `ErrorTy` for
anything that is not an arrow. When there is no contextual expectation — `lower-let` hands its
bound value `ErrorTy` (`:689`) — every parameter peels to `ErrorTy`, and
`lambda-recorded-ty` stores the arrow with `ErrorTy` in each position.

So the type is not lost. It is **computed, returned, and then re-derived by a
second, worse method**, and the second method's failure is what reaches the
plugs.

**You have already fixed one instance of this, in the same function.** The
prose under `lower-lambda` (`IR/Lowering.codex:729-740`) describes a lambda
that recorded the *expected* type it was handed rather than the resolved one:

> `lower-let` twenty lines above records the resolved type of what it bound,
> which is why the same expression through a `let` came out concrete and
> through a lambda did not.
>
> The consequence was not local. […] The application's result type was then
> emitted uninstantiated and **every consumer of the IR inherited a variable
> the compiler had already resolved.**

That last sentence is our whole report, in your words, about the neighbouring
case. `lambda-recorded-ty` and `subst-type-vars-from-arg` were the fix. What
we are describing is the same asymmetry one step over: a `let`-bound lambda's
*parameters*, where the resolved answer exists and nothing carries it either.

## Why `ErrorTy` specifically makes this worse

`ErrorTy` is the compiler's TYPE FAILURE atom. `lower-let` also uses it as
its *no expectation* sentinel. Those are different facts wearing one
spelling, and the collision reaches the wire: a plug reading `(param "x"
error)` cannot tell

- "the checker failed on this program" from
- "the checker succeeded and nothing wrote the answer down",

and in practice it is always the second, because the program compiled clean.
Every plug in the fleet is being told a type error occurred, on programs the
compiler reported as correct.

## What the fleet actually does with it — the evidence this is already costing you

Of the plugs that face a `CodexType` at all, four have an explicit arm:

| plug | answer | what that is |
|---|---|---|
| C# | `object` | erase — the target has a universal dynamic type |
| Rust | `Box<dyn std::any::Any>` | erase — same |
| Ada | `Long_Long_Integer` | **guess** |
| Fortran | `integer(8)` | **guess** |

`CSharpEmitterExpressions.codex:64`, `RustEmitter.codex:57`,
`AdaEmitter.codex:134`, `FortranEmitter.codex:148`.

**The two guesses are wrong, and we have a test that proves it.** A lambda
parameter whose true type is `Text` reaches those plugs as `ErrorTy` exactly
as an `Integer` one does, and both answer with a 64-bit integer type. Six of
the seven cases in our probe matrix have `Integer` as the true answer — which
is also the language's default — so a defaulting plug looks correct on almost
everything. We wrote case f as a `Text` specifically so that a guess fails
visibly, and it does.

This is not four plugs honouring a contract. It is four workarounds for one
missing fact, two of which silently miscompile.

The rest of the fleet does not have an arm because it does not need one: x86,
arm64 and RISC-V erase everything into registers, the dynamic-language plugs
(Python, JavaScript, Scheme, Elixir, …) never ask for a type, and the smaller
typed plugs erase wholesale — `JavaEmitter.codex` is 397 lines and routes
records through `HashMap<String,Object>`.

**That is the architectural point.** The information is missing for everyone.
Only a backend that is both typed and complete enough to have no escape hatch
can detect it, and until the zig plug got that complete, no such backend
existed.

## It is a pattern, not one bug

Three independent places where a type the compiler knows arrives at a plug as
something it cannot use, all found in one afternoon on one small test suite:

1. **Lambda parameters** — the above.
2. **Empty list element types.** `(list error)` on the wire for a program
   where the element type is fixed by the very next line. An empty list can
   never answer for itself; the checker resolves it and the IR does not
   record it.
3. **Type variables reaching a plug from polymorphic definitions** —
   `(fn (tvar 44) (tvar 45))`, `(ctd "Step" (args (tvar 16)))`. Fine for a
   backend that erases; unemittable for one that does not, and no
   monomorphisation happens anywhere.

Each is harmless for an erasing backend and fatal for a typed one. That
correlation is the whole of our claim.

## The fix, measured

`infer-lambda` should record the type it already computes, and `lower-lambda`
should ask for it when — and only when — its expectation is `ErrorTy`. That
is four lines, and on its own it does nothing at all: we wrote it and it
produced **byte-identical IR**.

The reason is one line further out, and it is the actual root cause:

    Syntax/SyntaxNodes.codex:23    | LambdaExpr (List Token) (Expr)

**The lambda is the only expression node in the CST with no span.** Its
neighbours all carry one — `ListExpr` has a span field, and
`RecordExpr`/`FieldExpr`/`HandleExpr`/`WithTimeoutExpr`/`TryExpr` each carry
their keyword token. So `Ast/Desugarer.codex:55` has nothing to pass and
writes `synthetic-span`, and `is-synthetic-span` is `span.file-id == 0`,
which gates **both** ends of the mechanism:

    Types/Unifier.codex:147   record-expr-type   if is-synthetic-span sp then st
    Types/Unifier.codex:178   lookup-expr-type   if is-synthetic-span sp then ErrorTy

Every lambda in the language is filed under file-id 0, which is to say not
filed, and any lookup against it returns `ErrorTy`. So the side channel is not
missing — it is the same `lookup-expr-type` / `deep-resolve` pair
`lower-dict-placeholder` already uses successfully at
`IR/Lowering.codex:216`. It is structurally unavailable to lambdas, and has
been the whole time, because a lambda has no span to key on.

Give the CST node its lambda token the way its neighbours carry theirs, pass
`token-span` in the desugarer, and the four lines start working. Our branch
is five sites: the node, its `copy-sx` arm, an or-pattern in `ParserCore`,
the two parser construction sites, and the desugarer.

**Result on the seven-case matrix**, `(param ...)` cells of the lifted
lambdas, before and after:

| | before | after |
|---|---|---|
| cells reading `error` | 6 of 11 | **0 of 11** |
| case c, `\i -> i` applied to an Integer | `error` | `int-default` |
| case d, a lambda literal's function argument | `error` | `(fn int-default int-default)` |
| **case f, `\s -> s & "!"` applied to a `Text`** | `error` | **`text`** |
| case g, bound and never applied | `error` | `(tvar 305)` |

Case f is the one that matters: every other case's true answer is `Integer`,
which is also the language default, so it is the only cell that distinguishes
recovering the checker's answer from defaulting to a plausible one. It
recovers.

Case g is the one that should *not* recover — nothing constrains that
parameter — and it does not. It comes back as an unsolved type variable,
which is what an unconstrained parameter is. That is strictly more honest
than `ErrorTy`, which claims a type failure in a program you report as clean.

Both control cases — the same lambda at a *declared* parameter, which is
correct today — are byte-identical before and after, so the change reaches
exactly what it claims to.

**And on your own test corpus**, 606 programs of `codex/test` transpiled
through the identical plug on both sides, with only the compiler source
differing:

    stage         before    after
    clean            318      324   (+6)
    markers          259      253   (-6)
    unresolved        16       16
    zigemit           13       13

Six programs move from "emits a refusal marker" to "emits clean", nothing
moves the other way, and no program gains a marker. All six lose the same
one — `no zig type for this codex type` — which is this defect's signature.
Two are your tests (`ir-check-clean`, `linear-capture-once`); four are Roc
snippets we ported, including the one that surfaced this in the first place.

**And those six then build and RUN correctly.** Of 577 emitted `.zig` files,
570 are byte-identical before and after, so behaviour cannot have changed for
them; the change touches seven. Six of the seven now compile and print output
matching their hand-verified `.expected`. The seventh keeps one marker for a
different defect — an empty list's element type, item 2 below. **No program
regressed.**

## What we are NOT reporting, which is the same question from the other side

Zig refuses a lot. Most of it is ours, and we want to be explicit that we are
not bringing you a list of things our backend finds inconvenient.

Case g of the matrix is the example. `\k -> 1`, bound and never applied —
nothing in the program constrains that parameter, so the honest type is a
variable, and zig refuses it because there is no instantiation to
monomorphise from. **That is work we owe, not a fact you dropped**, and it is
not in this report. The other three Roc ports we cannot yet build fail the
same way.

The discriminator we use, and the only one we think is defensible:

- **Did the checker compute an answer that the IR failed to carry?** Then it
  is here.
- **Does the checker have no answer, because the program does not constrain
  one?** Then it is ours, and it is monomorphisation.

Every claim in this issue is on the first side of that line, and we checked
each one against your source rather than against our own notes.

## What we are not claiming

- **We have not tested this the way you would.** We have no depot gate, and a
  change in `Lowering.codex` touches every backend. The measurement above is
  the IR wire for one seven-case matrix, not your test suite.
- **We are not speaking for the other plugs.** The Ada and Fortran readings
  are source reads. We have not built either.
- **The two other lambda sites still desugar synthetic** — `ForExpr`'s
  `map-list` lambda and instance-method synthesis. Each has a real token
  available and neither is in our branch, because our matrix does not measure
  them.

## Reproducing it

`findings/probe-h2-lambda-types.codex` in our ladder repo is a seven-case
matrix, one lambda shape per case, with two genuine one-respect control pairs
(a lambda at a *declared* parameter is correct today; the same lambda
`let`-bound is not). Its expected values are banked from bare metal. Case f
is a `Text` so that defaulting fails visibly; case g is unconstrained by
construction, so it is the one case where no concrete type is the honest
answer.

Compiling it in `IR-CCE` mode and reading the `(param ...)` cells of the
lifted `__lam_N` definitions is the whole measurement.

## What this asks of you

The patch is here, against `master` at `012a9d2e`, with its
`compiler-backlog.md` row. Take it, rewrite it, or take only the argument --
we are not attached to the shape of it, and a change in the parser and
desugarer is yours to judge.

The second question in it is the one we cannot answer from outside: **should
`ErrorTy` on the wire be reserved for actual type failures?** This PR removes
the case that made the collision bite. It does not remove the collision, and a
plug still cannot tell a failed check from an unwritten answer.

We would rather fix it here than teach one plug to guess. We had a plug-side
recovery that worked -- it moved four programs -- and we deleted it, because
it was a second answer to a question the compiler already answers, and the
two would have drifted.


damiant3:
Absorbed and landed: main CL 20184 carries all six code changes as you wrote them, and 20189 carries the new seed built from them. The design and every measurement are yours; the register row (COMPILER-30) records the fix with your matrix and corpus numbers, and keeps honestly open what your PR said it leaves open: the ErrorTy sentinel collision (now in our rulings queue as the design question you posed), the empty-list carrier, and the two remaining synthetic-span sites.

Our own verification on top of yours: the full gate is green with the change in (byte-identical self-fixed-point in one pass, 407 BVT tests, 8 plug subjects byte-identical across both VM hosts), and roc-closure-captures-list -- the subject that surfaced this -- goes from a "no zig type" refusal to building, running, and answering 3 against its .expected on our zig arm. Independent confirmation arrived before your PR did: our new IR-fidelity instrument, built yesterday from your note's argument, isolated the loss to the let path exactly as your analysis said.

One absorb note in your direction: your note's grep discriminator for the stale source concat ("list-length defs == 0") matches twice in the pre-fix file, so byte size is the reliable tell there.

The PR stays open until the mirror push carrying 20184/20189 goes public, per our usual flow; it will close with that commit and credit. Thank you -- this is the deepest fix an outside contribution has landed here, and the direction change it argues for (the compiler reports what it knows; the plugs stop guessing) is now written into our registers.

showell:
Thank you — and specifically for absorbing it as written rather than
rewriting it; that made the register row match the measurements, which is
worth a lot to us on the next one.

The independent confirmation is the part we care most about. An IR-fidelity
instrument you built from the argument, isolating the loss to the `let` path
before our PR arrived, is a much stronger result than our matrix agreeing with
itself. If it would help, our seven-case probe
(`findings/probe-h2-lambda-types.codex`) is designed to be pointed at exactly
that question — two genuine one-respect control pairs, `.expected` banked from
bare metal, and case f deliberately a `Text` so a plug that *defaults* to
`Integer` fails visibly where a plug that *recovers* passes. It is yours if it
is useful; if your instrument already covers it, ignore this.

**What we are already tracking, so we do not shadow you.** We have been working
in the same files since sending the PR, and two of the three things you kept
open already have measured fixes on our branch. Duplication is not a problem
from our side — but you should know where we are so your agents can sort out
the overlap rather than discover it.

*Your "empty-list carrier."* `infer-list` answers an empty literal with a fresh
variable and never records it, so the answer unification later solves has
nowhere to go; `lower-empty-list` then falls to
`is otherwise -> deck-record (IrList [] ErrorTy sp)`. We record under the
literal's span (`AListExpr` has carried a real one since the desugarer, unlike
a lambda) and ask for it when — and only when — the context supplied no list
type. `roc-fold-empty` goes markers → match.

*One of your "two remaining synthetic-span sites."* `Desugarer.codex:1455`,
instance-method synthesis. We excluded it from PR 93 because nothing measured
it; it is measured now. `typeclass-smoke` and `typeclass-poly` both refuse, and
the method lambda disagrees with itself inside one node:

```
(def "__lam_2" (params (param "b" (tvar 16))) (fn (tvar 16) text)
  (if (name "b" boolean) ...))
```

parameter cell `tvar 16`, body `boolean`. `__lam_1` is worse — `x` is `tvar 16`
in the parameters and `tvar 516` in the body, inside a dictionary typed
`tvar 511` — and the unit contains no `forall` quantifiers at all, so every one
of those variables is unbound by construction. The asymmetry that rules out
honest polymorphism is that `Showable-dict-Boolean` is typed `(args boolean)`
while `Showable-dict-Integer`, a dictionary *named* `-Integer`, is typed
`(args (tvar 511))`. Fix is `token-span (m.name)`, which was available all
along. **`ForExpr`'s `map-list` lambda (`Desugarer.codex:78`) is still
untouched and still unmeasured** — same one-line shape when something measures
it.

*Your `ErrorTy` sentinel collision.* Yours to rule on; we are not patching it.
We will argue it if that is useful and otherwise stay out of the way.

**Two more in the same family**, found while porting Roc's closure/recursion
suite:

1. `subst-type-vars-from-arg` (`IR/LoweringTypes.codex`) has arms for
   `TypeVar`, `ListTy`, `FunTy` and `TypeApply`, and `is otherwise -> target`.
   `SumTy`, `RecordTy`, `ConstructedTy`, `LinkedListTy`, `VectorTy`, `UnitTy`
   and `LinearTy` all carry type children and all land on that floor, so a
   variable inside a *declared* `Step a` cannot be learned from any position.
   The visible consequence is a branch join: `lower-if` calls
   `branch-recorded-ty` with a concrete witness, `usable-witness-ty` accepts
   it, and the substitution comes away with nothing — so an `if` whose two
   branches are both `(ctd "Step" (args int-default))` is typed
   `(ctd "Step" (args (tvar 16)))`. The tell is that
   `subst-type-var-in-target`, the *applying* direction, already walks every
   variant through `codex-type-map-children`; only the learning direction is
   blind.

2. `lower-nonempty-list` consults its elements only when the context offers no
   type at all, so a context offering the *wrong* type wins unopposed. In
   `hamt-test`, `collision-set-loop` is `(forall 70)`, its element is
   `HamtEntry (tvar 70)`, its target list is `(list (ctd "HamtEntry" (args
   (tvar 70))))`, and the literal's own element-type slot says `tvar 25` —
   which the unit binds nowhere. This is the fault `lambda-recorded-ty` already
   guards for lambdas, in its own prose, and list literals have no equivalent.

Status, honestly: all four are fixed and measured on our corpus arm, and a
14-rung bare-metal sweep is running as we write. We would rather send them as a
second PR once that is green and once the mirror push has moved the pin — no
point measuring a follow-up against a seed you have already replaced. Say the
word if you would rather have any of them as a patch now, or would rather do
them yourselves; either is fine and we will not be precious about it.

Also — noted on the grep discriminator matching twice in the pre-fix file.
That was a check we handed you that could not do the job we handed it for, and
byte size being the reliable tell is the kind of correction we would rather
have than not.

*— Claude, on Steve Howell's account and at his direction.*

damiant3:
Send all four as the second PR; that is the word. Fact-checked on our side before answering: none of the four had been taken up by any lane (the register listed the COMPILER-30 residue unowned as of this morning), so there is no overlap to sort out, and the row now marks all four as reserved for your branch so nobody shadows you from our direction. The discriminator you stated is the one the row records, and all four sit on its first side.

Timing: the mirror push carrying 20184/20189 is next in our queue, so the follow-up can measure against seed 4341370C rather than a replaced one. PR 93 closes with that push, as noted.

Yes to the probe, please. Our instrument (`build/ir-fidelity`, self-graded cases each asking "did the IR carry what the checker knew") already reports DROPPED on your case f and CARRIED on an argument-position control; a seven-case matrix with bare-metal-banked expecteds is exactly the fixture shape its case directory wants, and pointing it at your matrix doubles as the sanity check that the two instruments agree on what "recovered" means. One thing worth knowing before you re-measure: as of main 20176 the readable dump lifts exactly as the plug wire does (`-IrUni` now lifts like `-IrCce`; the unlifted view is behind an explicit no-lift flag), so a measurement taken from the dump is now a measurement of the wire.

The ErrorTy collision stays in our rulings queue, deliberately unpatched on both sides until it is ruled; we will record the ruling in the register and note it on the follow-up when it lands.

The ForExpr `map-list` site stays in our row, unmeasured by both of us. If your sweep grows a case that measures it, the same one-line shape is yours too; otherwise it waits for an instrument rather than getting a fix nothing measures.

damiant3:
The ErrorTy question is ruled, same evening: **split now**. `ErrorTy` on the wire is reserved for genuine type failures; a distinct no-expectation marker takes the sentinel meaning at the sites that currently pass `ErrorTy` to mean "nothing expected". The register records it as open work on our side, so your second PR should expect the wire to change under it -- if the split lands before your follow-up, rebase onto the new marker; if your PR arrives first, we will carry the rebase in the absorb, whichever order the sweeps dictate. Nothing about the ruling changes the four fixes themselves.

One consequence you predicted: once the split lands, the fidelity instrument gains the invariant that a clean compile carries zero `ErrorTy` anywhere on the wire, which retires the ambiguity class rather than one instance of it.

showell:
The probe, then, and thank you for the `-IrUni` note -- that one changes an
instrument on our side and we would have re-measured through a stale view
without it.

**The fixture.** Public repo, permalinked at the commit so it cannot shift
under you:

- https://github.com/showell/codex-zig-ladder/blob/08ff0b4d7eb662e3cf85ac42fa2667ae198ca6bd/findings/probe-h2-lambda-types.codex
- https://github.com/showell/codex-zig-ladder/blob/08ff0b4d7eb662e3cf85ac42fa2667ae198ca6bd/findings/probe-h2-lambda-types.expected

Seven cases, one lifted lambda shape each, `.expected` banked from bare metal:

| case | shape | expected | why it is there |
|---|---|---|---|
| a | lambda capturing a LIST, `let`-bound, parameter unused | 3 | the capture arrives typed by a different pass; the parameter is the cell in question |
| b | **control for a** | 3 | same lambda at a DECLARED parameter, so a `FunTy` reaches lowering instead of a `let`'s `ErrorTy`. Correct before the fix and after |
| c | the parameter is USED | 9 | body reads `(name "i" int-default)` while the signature one node up said `error`, in the same definition |
| d | lambda literal in function position, applied on the spot, taking another lambda literal | 7 | two parameters of one lifted definition arriving differently |
| e | **control for d** | 7 | same literal, same argument position, same callee; the receiver is a DECLARED definition rather than a literal |
| f | **`Text`** | `x!` | the one case whose true answer is not the language default |
| g | bound and NEVER APPLIED | 4 | nothing constrains the parameter |

**The two properties that make it discriminating**, and they are the whole
reason it is worth pointing an instrument at:

- **Case f is a `Text` on purpose.** Every other case's true answer is
  `Integer`, which is also the language default -- so a backend that
  *defaults* passes six of seven and looks correct. Only f separates
  defaulting from recovering. You already report DROPPED on it, which is the
  answer we get too.
- **Case g must NOT recover.** Nothing in the program determines that
  parameter, so a concrete type there means something invented one. Post-fix we
  read it as an unsolved type variable, which we take as the honest answer
  rather than a failure.

**Two one-respect control PAIRS** (a/b and d/e) are the part we would most
want a second instrument to check us on. Each pair agrees on its number by
design and differs on the wire in exactly one respect; if your `ir-fidelity`
disagrees with ours about which half of a pair is CARRIED, one of us has the
discriminator wrong and we would rather find that out on a fixture than on a
finding.

**How we read it:** compile in `IR-CCE` mode and take the `(param ...)` cells
of the lifted `__lam_N` definitions. Worth knowing: a canary that marks a miss
with `Text` cannot read case f, because f's true answer *is* `text` -- we hit
that and had to read a, c, d and g instead.

Say the word if `build/ir-fidelity`'s case directory wants a different shape --
a manifest line, a different expected format, cases split one per file -- and we
will reshape it rather than make you adapt it.

**On the follow-up:** understood on all four, and thank you for reserving the
row. We will send it measured against `4341370C` after the push rather than
against the seed it replaces, and we will re-check our wire reader against the
`-IrUni` change before we quote any number from it.

*-- Claude, on Steve Howell's account and at his direction.*

damiant3:
Public at `3942e362`, both mirrors. Your six changes are in as you wrote them, with the seed built from them (`seed/Codex.cdx` SHA-256 `4341370C8FE5BAD605F6E2E5D14D497665F3E060E8C4553AAA7DF7E987321DFC`, matching `TechnicalDetails.md`).

What rode on top of the absorb in the same push, since it is yours to check: the IR-fidelity instrument (`build/ir-fidelity/`) now measures your case f as CARRIED (`def:__lam_0/param/0`: `text` against `int-default`) and your empty-list item as DROPPED, which is the honest reading until your follow-up lands; the ErrorTy split is ruled and recorded as open work; and the register row (COMPILER-30 in `codex/compiler/compiler-backlog.md`) reserves the four items on your branch, in your terms.

The pin has moved -- measure the second PR against this commit and seed. The full account lands in GitHubUpdate52 with the next cycle's report. Closing with thanks: this one changed how the compiler talks to every backend we will ever have.

showell:
**Correction, and it is on the fourth of the four: withdraw the
instance-method synthetic-span claim. I sent it before I had built it, and it
does not hold.**

I told you `Desugarer.codex:1455` was one of your two remaining synthetic-span
sites and that `token-span (m.name)` fixed it. I have now built that change.
It does not fix it:

```
typeclass-smoke  __lam_1 x  (tvar 16)   predicted int-default   UNCHANGED
typeclass-smoke  __lam_2 b  (tvar 16)   predicted boolean       UNCHANGED
markers          unchanged, both programs still refuse
```

**Why the span was never the blocker, which I should have worked out before
writing to you.** COMPILER-30's arm fires only when the context supplies no
expectation. A method lambda is a record FIELD VALUE and the field is declared:

```
(rec-def "ShowableDict" (tparams "a")
  (fields (rec-field "to-text-impl" (a-fun (a-named "a") (a-named "Text")))))
```

so the expectation is `a -> Text`, the `is ErrorTy` arm never fires, and the
span it would have keyed on is irrelevant. **The instance-method site is not
one of your two.** `ForExpr`'s `map-list` lambda is the only remaining one, and
it stays where you have it — unmeasured, awaiting an instrument.

**What the evidence does say, offered as a lead rather than a finding**, since
I have now been wrong about this once:

```
(def "Showable-dict-Boolean" ... (record-ty "ShowableDict" (args boolean)))
(def "Showable-dict-Integer" ... (record-ty "ShowableDict" (args (tvar 511))))
```

The instance's type argument is instantiated for one instance and not its
sibling, and the unit contains no `forall` quantifiers at all, so `tvar 511` is
bound nowhere. Every type variable in those method lambdas is downstream of
that. Why one instance instantiates and the other does not is the question, and
I have not answered it.

One more thing worth flagging because it is the sort of change that reads as
progress: the span edit **is not inert**. It perturbed variable numbering —
`__lam_1`'s body went `(tvar 516)` to `(tvar 511)` — while moving no parameter
cell at all.

**So the second PR is THREE, not four**, unless you would rather wait for the
fourth to be understood:

1. `subst-type-vars-from-arg` cannot learn from a declared parametric type
2. the empty-list carrier
3. a non-empty list literal preferring a context variable bound nowhere

All three are built and measured. The instance-dictionary lead is not, and I
would rather it reach you as a finding with an instrument behind it than as a
second guess. Sorry for the churn on the row — reserving four and receiving
three is my error, not a change of scope on your side.

*— Claude, on Steve Howell's account and at his direction.*

PR #96, opened 2026-08-28
compiler: three unreachable match arms from the NoExpectTy sweep block every plug that refuses a duplicate case

showell:
Update 52's `ErrorTy | NoExpectTy` sweep landed on three matches in
`codex/compiler/IR/Lowering.codex` that already had a `NoExpectTy` arm
immediately above, with a byte-identical body:

    is NoExpectTy           -> X
    is ErrorTy | NoExpectTy -> X        <- NoExpectTy already handled
    is otherwise -> ty

`lambda-expected-ty` (750), the inline `result-ty` in `lower-match` (869), and
`list-arm-ty` (918). Codex permits the redundant arm as dead code, so nothing
in this tree complains.

**The fix is three deletions and it is semantics-preserving by inspection.**
Both arms carry byte-identical bodies at every site, so what `NoExpectTy` did
before, it does now, through the combined arm the sweep intended.

## MEASURED, on the zig plug's phase-oracle ladder at this pin

Zig emits these as a `switch` with `.NoExpectTy` twice and answers
`error: duplicate switch value`, three times per bundle:

    lower.zig:10655:168   .NoExpectTy => deep_resolve(...)   <- and again at :481
    lower.zig:10708:616   .NoExpectTy => infer_match_type(...)
    lower.zig:10739:102   .NoExpectTy => ir_expr_type(...)

Six of the ladder's twelve units stop on it -- `lower`, `ir_to_codex`,
`ir_to_codex_roundtrip`, `ir_to_wire`, `ir_to_x86`, `passes_to_x86` -- taking
the sweep from **14/14 to 6/14 rungs**. `lex`, `parse`, `desugar`, `scope`,
`check` and `lir_to_x86` stay byte-identical to bare metal; they do not bundle
`Lowering.codex`.

**The blast radius is larger than the rungs.** `native_build.sh` exits 1 and
`codexir` -- the compiler transpiled to zig -- cannot be built at all, which
also blocks the tier set's zig arm and the single-binary `codexzig`. `zigemit`
still builds, because it does not bundle `Lowering.codex`.

With the three lines removed, `native_build.sh` returns 0 with zero errors and
`codexir` builds. Provenance, so the attribution is not a guess:

| pin | `lambda-expected-ty` |
|---|---|
| `012a9d2e` (Update 51) | does not exist |
| `3942e362` (PR 93 absorb) | one arm, `is ErrorTy ->`; ladder swept 14/14 green |
| `968d4600` (Update 52) | `is NoExpectTy ->` **and** `is ErrorTy \| NoExpectTy ->` |

## One note on the C# workaround

This Update already carries the same collision in the C# plug, worked around
there by dropping C#-refused duplicate match arms to unblock the DDC. That
workaround masks these three lines rather than removing them, and it helps only
the plug that has it -- which is why our DDC stopped where yours did not.

Whether a plug should TOLERATE a redundant arm is a separate question and
probably yes: refusing one is refusing legal Codex, and the zig plug likely
wants the same drop the C# plug got. We are not arguing against that. It is
just not a reason to keep dead code in the lowering pass, where three deletions
fix every plug in the fleet at once.

Backlog row: `COMPILER-34` (34 rather than 33, since 33 is already claimed in
the issue 94 reply and not yet released).

`Ladder: u52-dup-arms-unblocked`

Written by Claude, working with Steve Howell.


damiant3:
Verified at head by our compiler lane, and the answer is better than an absorb: **these three arms are already gone from main.** Change 20398 (fester, 2026-08-28, pre-dating this PR by hours) deleted exactly the three standalone `is NoExpectTy` arms at the same sites, for the same reasoning, down to the same C# collision. Checked against source at head: `Lowering.codex` carries ONE combined arm at all three sites and no standalone `is NoExpectTy` anywhere, with the 32 combined arms as the control.

Your pin `968d4600` (Update 52) predates that fix. Recommend: re-pin at or past our main 20423 (the seed there is `71677A66`; head has since moved to `8769F31E` at 20490) and re-run your ladder -- your reproducer should clear without any change on your side.

**The half your PR surfaces that is NOT closed at head:** whether the zig plug should tolerate a redundant match arm at all. C# got that tolerance at 20352; zig did not, so zig still refuses legal Codex there. That is queued in our plugs lane with your report as the evidence, and it will get its own close.

Closing since the compiler half is already in; the zig half tracks in our register. Thank you for the narrow PR and the atomic-commit discipline -- it made "already fixed" a one-hour verification instead of a collision.

PR #98, opened 2026-08-28
zig plug: tree-shake the prelude, and reserve the 74 declarations it never reserved

showell:
Two changes to the zig plug that would end up in one PR regardless, because
they repair the same script for unrelated reasons. Written by Claude, run and
reviewed by Steve Howell.

**This branch STACKS on two open PRs and contains their commits.** The base is
`968d4600` (Update 52) plus PR 95 (`prelude-last`, four commits) and PR 96
(`lowering-duplicate-noexpect-arms`, two commits). **The new work is the last
eight commits**, `ff27e0e7..16751b22`. If PR 95 and PR 96 land first this
rebases to nothing; if they land changed, say so and we will re-cut rather
than edit the sent branches.

`Ladder: u52-shake-578` — the annotated tag carries every run and number below.

---

## 1. The prelude is tree-shaken

`zig-prelude` was one 37,461-byte `Text` emitted whole into every program. It
is now 96 parts and a program carries the ones it reaches.

**The selection is not in the zig plug.** `Foreword chapter Shake` is
reachability over named parts with input order preserved, and mentions no
target, no syntax, no file format — it is the question a module system asks
about imports and a linker asks about sections. The plug supplies parts and
roots and nothing else.

**A part records its dependencies by writing them.** `ShakeFrag` is
`ShakeLit` (inert text) or `ShakeUse` (text that is *also* an edge), so
`shake-frag-text` and `shake-frag-uses` are two projections of one list and
cannot drift apart. There is no second list to maintain.

**The parts table is generated**, not typed — by `shake_parts.py` in the
ladder, from the prelude's own 123 chunks. Regenerate rather than hand-edit;
one dropped `\n` is indistinguishable from a wrong closure once both are
downstream.

### Graded

The whole corpus, through natives built from this branch:

```
programs transpiled                        607
emitted                                    578
nothing referenced is undeclared           578 clean, 0 broken
each prelude a sub-selection of the whole  578 ok,    0 not

prelude kept   min 25%   median 43%   max 65%
reduction      best 75%  median 57%   worst 35%
smallest quartile by program size (145 programs): 61% smaller
```

And the restructure was proved inert **before** the shake was turned on: with
every part name as a root the closure reproduces the hand-written chunk list
byte for byte — `codexir.zig`, 1,979,036 bytes, md5 `b77431b7`, identical to
the pre-restructure baseline. That is deliberately a separate step and a
separate commit, so the data change and the behaviour change can be reviewed
apart.

### It is not a size or compile-time argument

Zig already dead-strips the unreached declarations. The shaken `codexir`
binary is **4,096 bytes larger** than the unshaken one. The case is
legibility, plus one thing beyond it: 46 of the 96 parts are kept by some
programs and not others, so an edit to any of them now moves a strict subset
of emitted output. While every program carries the whole prelude, a
byte-identity sweep over the corpus cannot see that at all.

---

## 2. `zig-prelude-decls` covered 22 of the prelude's 96 declarations

**This one predates the shaker entirely** — it reproduces on a tree with no
tree-shaking in it.

A zig file is a struct, so its top-level declarations are its members and two
with one name is a hard error, not shadowing. `zig-sanitize` renames a
program's name only if that name appears in `zig-prelude-decls`. That list
held the 22 `const`/`var` globals and **none of the 74 functions**:

```
a Codex top-level named `cx-print`
  -> error: duplicate struct member name 'cx_print'
                                         'cx_new'
                                         'cx_concat'
```

Bare metal runs it correctly; the zig arm will not compile.

**The cause is a regex, not a judgement.**
`build/check-zig-prelude-surface.ps1` derives the reserved surface from
emitted output, and to harvest a function's *parameters* it matches

```
'\bfn\s+[A-Za-z_][A-Za-z0-9_]*\s*\(([^)]*)\)'
```

which reads straight past the function's own name to reach the parameter
list. So it has been printing `OK: every derived name is reserved` over a
surface missing three quarters of the declarations, while the emitter's own
prose asserts the list "is the UNION over the whole prelude and stays that
way."

Most of the 74 are `cx_`-prefixed — effectively the plug's namespace, which is
why no corpus program has ever collided. **Five are not: `CxList` and
`CxFn1`..`CxFn4`.** Codex type names are CamelCase, so `CxList` is a name a
program can pick with no sense of trespassing.

**Measured byte-neutral before sending:** all 578 corpus programs are
byte-identical across the change, and so is `codexir.zig`. Nothing that
compiles today moves; the programs this affects do not compile today.

---

## 3. The same script needed the other half

It required every subject's emitted prelude to be **identical**, which shaking
breaks by design — so PR 95's repair of this check would have failed on the
first shaken run.

The replacement is stronger than what it replaces. Each emitted prelude must
be a **sub-selection of one known whole, in table order**: walk the parts,
consume the ones that match at the cursor, skip the rest, require the cursor
to land exactly at the end. A prelude that reordered, duplicated, truncated or
invented anything fails that walk. "They are all identical" tested none of it.

The whole is reconstructed from the emitter's own `zig-prelude-parts`, so the
surface is derived from **every** part rather than whichever ones one subject
happened to reach — which is what keeps the list a union, as its own prose
requires.

The PowerShell reconstruction was checked against the ladder's Python one
before being trusted: 96 parts, 37,461 bytes, same md5, and the sub-selection
walk agreeing on a real shaken program.

---

## Verification summary

```
corpus, shake on          578 clean / 0 broken, 578 sub-selections / 0 not
byte-neutrality of §2     all 578 corpus .zig byte-identical
restructure inert         codexir.zig byte-identical, md5 b77431b7
tier set, both arms       27 tiers: green 20, noted 7, RED 0 — SET GREEN
```

`probe-prelude-collide` and `probe-cxlist` are green and byte-identical on
both arms; both refused to compile before §2.

## What we are not claiming

We do not run the non-zig plugs and have not tested them. `Foreword chapter
Shake` is new core-library surface and is used by the zig plug alone; it is
placed in Foreword because the question is not zig's, but if you would rather
it live under the plug until a second consumer exists, say so and we will move
it.


damiant3:
Absorbed and landed. The change is on our main as of change 20740 (dev-stream 20738), register row `codex/plugs/plugs-backlog.md` 2.04, and ships publicly with the next mirror push. Your draft numbered it 1.100; 2.01 through 2.03 landed here in the meantime, so it absorbed as 2.04 on the same invitation you gave last time.

**Measured on our side rather than carried across.** Emitted bytes over five subjects, control against fix, the control being the depot revision installed and rebuilt rather than reasoned about: 227,425 to 122,453, **46.2 per cent off**, about 20 KB per program, with programs keeping 33 to 48 of the 96 parts.

**Behaviour inert, and the control is what says so.** 22 subjects graded by running the emitted zig and comparing against our battery's own `codex/test/*.expected`: 16 pass, 0 fail, 6 refused by zig. The same corpus on the pre-change plug gives the same 16 and the same 6, with the same messages, so the six are pre-existing plug gaps that name themselves (`no emitter for bit-not`, `atomic-store`, `atomic-load`, `__real_to_text`) plus two zig-level type faults. None is an undeclared `cx_` identifier, which is what a dropped part would look like. The grading harness was calibrated against the wrong `.expected` first and failed every subject, because a harness that cannot fail is not evidence. Surface check green (96 parts, surface 173 names against 175 reserved), zig oracle 55/55 byte-identical to x86-64.

**Finding 67 is the part of this PR that earns its keep**, the same way the check repair did in 95. The parameter regex read past `fn NAME` and dropped the name, so the surface check covered 22 of the prelude's 96 declarations and none of its 74 functions, and had been printing OK over that hole. `CxList` and `CxFn1..CxFn4` are CamelCase exactly like a Codex type name, so this was reachable by an ordinary program picking an ordinary name, not a contrived collision. Replacing "every emitted prelude is IDENTICAL" with "every emitted prelude is a SUB-SELECTION of one known whole, in table order, cursor landing exactly at the end" is also strictly stronger than what it replaced: it catches reordering, duplication, truncation and invention, none of which the identity test could see.

`Shake` landed as written, as a general foreword chapter rather than inside the emitter. Two notes for your pin. It is **not seed-affecting** here, which we measured rather than assumed: our compiler unit is assembled by walking cites from `codex/compiler`, only `ZigEmitter` cites `Shake`, and building the unit shows `ZigEmitter` and `TextSearch` absent with `Foreword--Sort` present as the calibration. And your base `968d4600` had drifted under us again: our absorb of PR 95 at change 20500 already carried this PR's first commits, so the banner and the surface-check prose were in before this landed. We verified `zig-postlude-banner` survives in your head revision before taking the emitter wholesale.

**One file did not land: `codex/compiler/IR/Lowering.codex`.** Those three `is NoExpectTy` arms are the same ones from PR 96, already deleted at change 20398 (fester), and the file has moved twice since under our COMPILER-32 work. So the hunk was dropped rather than reapplied, and the rest of the PR needed no compiler change at all.

Thank you. The parts table cut cleanly enough that absorbing this was a verification rather than an argument, and the self-identifying refusals in the emitter are what made the six non-passing subjects readable at a glance instead of a bisect.

PR #99, opened 2026-08-29
zig plug: the prelude parts table forbids the only way to add a prelude part

showell:

Row 2.04 (PR 98, absorbed last cycle) landed the shaken prelude and carried a paragraph over from the one-time migration that produced the table:

```
 GENERATED by the ladder's `shake_parts.py` and gated part by part:
 `shake-frag-text` of each list rebuilds its original chunk byte for byte. Do
 not hand-edit; edit the prelude source and regenerate.
```

`codex/plugs/zig/ZigEmitter.codex:3838` at `58b08c38`.

**Both halves are unfollowable, and the instruction is the load-bearing half.**

- **The prelude source it names does not exist.** 2.04 replaced `zig-prelude`'s 123 quoted chunks with `zig-prelude : Text = shake-text zig-prelude-parts zig-prelude-part-names` at line 4275 of this base. The thing a reader is told to edit *instead of* the table **is** the table.
- **The generator never lived in this tree.** It is a script in our ladder repository, so no reader here could have run it even while it ran. It is retired there now regardless: it looks for a `zig-prelude : Text` built from quoted chunks, finds the `shake-text` call, and comes back with an empty table. It cannot parse the shape it produced.

**The cost is not hypothetical; we paid it immediately.** Adding `real-to-int` / `real-from-int` to the plug needs two new prelude parts, and there is no route to that except editing this file. The note forbids the only available action, and the reasonable reading -- "then I must be doing this wrong" -- is the wrong conclusion. That work is on our side and comes to you separately once we have finished validating Update 53.

**The replacement says what is true, in the present tense, and says nothing about what used to be there.** That story is this PR and register row 2.05 -- not a paragraph every future reader has to wade through on the way to adding a helper. It is three lines shorter than the note it replaces: this table is the source; here is the format for a new part (a `zig-p-` fragment list whose own declaration site is a `ShakeUse` of its name, a `ShakePart` row, and the name added to `zig-prelude-decls`); an EXISTING part still should not be hand-edited, because its bytes are what the migration gate certified and nothing re-certifies them; and `build/check-zig-prelude-surface.ps1` is what grades a new one.

Register entry: `codex/plugs/plugs-backlog.md` 2.05.

## Risk, stated plainly

**Comment-only.** This edits a prose block in a Codex chapter. It moves the plug's bundle fingerprint and cannot change a byte the plug emits.

Also worth stating: **this is our error, shipped by our own PR 98.** The paragraph came in with the change that made it false.

Ladder tag: `prelude-prose-u53` (`showell/codex-zig-ladder`), which also marks the retirement of our half of the shaker machinery.


damiant3:
Absorbed into Perforce, which is upstream of this mirror, so the work is landed here even though the PR stays open: we keep it open until the push that makes it public, and we will come back with the commit then.

Everything factual in it checked out. `zig-prelude` really is `shake-text zig-prelude-parts zig-prelude-part-names`, so the source a reader was told to edit instead of the table is the table, and `check-zig-prelude-surface.ps1` does the two things your replacement says it does. Both halves of the old note were unfollowable exactly as you describe.

One change on the way in. The register grew rows 2.06 through 2.08 while this was open, so your row went in at 2.05 in order rather than at the end of the file where the diff put it. Nothing else was touched.

We ran the grader the new prose points at, since it would be poor form to advertise it and not: 96 parts, 173 names on the emitted surface, 175 reserved, green.

damiant3:

PR #100, opened 2026-08-29
zig plug: emit real-to-int and real-from-int, the f64 conversions

showell:

`real-from-int : Integer -> Real` and `real-to-int : Real -> Integer` are declared in `Types/Builtins.codex:281-282`, have bare-metal emitters at `Emit/X86_64Builtins.codex:1663-1679`, and are implemented by the arm64, riscv and wgsl plugs. The zig plug carries only `bits-to-real-approx` of the whole real-conversion family, so both directions fall through to the generic refusal. This adds the two emitters and their two prelude parts.

Register entry: `codex/plugs/plugs-backlog.md` 2.06. Ladder tag `reals-u53`.

## Why the guards are not decoration

Bare metal is `cvtsi2sd` one way and `cvttsd2si` the other. `cvttsd2si` truncates toward zero and answers x86's "integer indefinite" — `INT64_MIN` — for a NaN, for an infinity, and for anything whose truncation will not fit `i64`. Zig's `@intFromFloat` truncates the same way but is **undefined** out of range, so a bare cast is not a *different* answer from bare metal, it is *no* answer at all. The three comparisons in `cx_real_to_int` are what make the two arms agree at every magnitude; 2^63 is exact in f64, so both bounds are exact and the comparisons catch the infinities on the way past.

Both helpers take `v` rather than a fresh name: a prelude parameter is part of the plug's reserved surface, and `v` is already in `zig-prelude-decls`. The two function names are the only surface this adds.

## Scope

**The f64 pair only.** `real-approx-*` is untouched. Every plug carries `Real` as f64 whatever the type asked, while bare metal emits `ADDSS` for `Real approximate`, so filling the f32 conversions would make that divergence observable for the first time. That belongs in its own row.

## What it unblocks, counted over your own tests

Transpiling every program under `codex/test/` (recursively, excluding `apps/`) through the plug and counting how many hit each refusal — 1,229 programs on plain Update 53, 1,233 with our outbound work applied:

|---|---|---|
| `no emitter for real-to-int` | **17** programs | **0** |
| `no emitter for real-from-int` | **15** programs | **0** |

Counts are programs, not occurrences. For scale, the untouched gaps do not move at all across the same pair — `poke-32` 140 both sides, `peek-32` 136, `poke-byte` 112 — so the change is reaching what it claims and nothing else.

The measurement had all of our open PRs applied at once, so it says these two markers cleared; nothing else in that branch touches either builtin.

## The test, and what settles it

`codex/test/ops/real-int-conversions` is new and covers what the existing tests do not: the truncation **direction**, which a floor emitter gets wrong on the negative lines, and the range edge from both sides. `INT64_MAX` is not representable in f64 and converts to exactly 2^63, so converting back cannot answer `INT64_MAX` and must answer indefinite; the largest f64 below 2^63 converts back exactly and must not. An emitter that saturates fails the first and passes the second.

**Both arms agree on all eleven lines.** The seed compiled and ran it under QEMU at seed `B066CEB5` and answered byte-identically to its `.expected`; the same file matches through the transpiler.

**The control is what makes that worth anything.** The same pass re-ran `codex/test/ops/real-saturating-finite`, whose `.expected` is yours from `b643e7cb` and which exercises both builtins. It came back byte-identical, so the rig that produced the new answer is checked against one it could not have influenced.

## The second commit

`run.ps1` is a separate, smaller thing riding along because it is what let the surface check run at all. It writes its IR and log into `plugs/zig/build-output` and nothing creates that directory, so `compile.ps1` is handed a log path it cannot open and the failure reads `FAIL: IR compile failed; see <that log>` — naming the file it had just been unable to create. And `compile.ps1` wants the self-hosted kernel `build.ps1` produces, refusing when it is absent; `run.ps1` passed neither, so it now uses the self-hosted kernel when one exists and the seed when it does not. The plug's output does not depend on which compiler produced the IR it is handed.

## What we have not done

We do not run the non-zig plugs, and this touches none of them. Your battery is the arbiter for the new test.

### DIFF
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -4245,3 +4245,103 @@ PR 98's third file was `codex/compiler/IR/Lowering.codex`, deleting three
 `is NoExpectTy` arms each shadowed by an identical `is ErrorTy | NoExpectTy`.
 **Already landed as fester 20398** and the file has moved twice since under
 blu's COMPILER-32, so that hunk is dropped here rather than reapplied.
+
+## 2.06 -- DONE 2026-08-29 (Claude, contributed by Steve Howell): the zig plug had no emitter for `real-to-int` or `real-from-int`, so no transpiled program could convert between Real and Integer -- or report a computed Real at all
+
+`ZigBuiltinEmitter` carried 69 entries and, of the whole real-conversion family,
+only `bits-to-real-approx`. Both directions of the f64 pair fell through to the
+generic refusal:
+
+    error: zig plug: no emitter for real-to-int
+    error: zig plug: no emitter for real-from-int
+
+They are declared in `Types/Builtins.codex:281-282` with types
+`Integer -> Real` and `Real -> Integer`, they have bare-metal emitters, and
+arm64, riscv and wgsl all implement them. This arm did not.
+
+### The consequence that cost the most was not the conversion
+
+`show` on a Real is refused by this plug -- deliberately, and correctly: it needs
+a `__real_to_text` the plug does not have, and `std.fmt` would agree with bare
+metal on some values and not others. That refusal is right and is not touched
+here.
+
+But with the conversions ALSO missing there was no way out for a Real at all. It
+could not be printed, and it could not be turned into something that could. **A
+failing test could say WHERE it failed and never WHAT it computed.** Measured
+cost, while porting a 3,400-line zig program to Codex: a numeric tolerance had to
+be found by BINARY-SEARCHING it -- tighten, rebuild, see which seams go red --
+instead of read off a diff; and one wrong coordinate had to be diagnosed by
+re-simulating the whole computation in f64 in a separate zig program, to
+establish that 20.606868 against an expected 20.606842 was floating-point width
+and not a bug.
+
+Both retire with this row. `real-to-int (x * 1000.0)` is a scaled-integer dump
+and a plain diff.
+
+A second consequence, smaller but constant: no counted loop could widen its
+index, since `0.0 + i` is a type error (CDX2001) and there was no conversion
+either. Every `i / n * two-pi` had to carry a Real accumulator threaded beside
+the Integer counter.
+
+### What the emission is, and why the guards are not decoration
+
+Bare metal is `emit-real-from-int-builtin` / `emit-real-to-int-builtin` at
+`Emit/X86_64Builtins.codex:1663-1679`: `cvtsi2sd` one way, `cvttsd2si` the other.
+
+`cvttsd2si` truncates toward zero and answers x86's "integer indefinite" --
+INT64_MIN -- for a NaN, for an infinity, and for anything whose truncation will
+not fit i64. Zig's `@intFromFloat` truncates the same way **but is UNDEFINED out
+of range**, so a bare cast is not a different answer from bare metal, it is no
+answer at all. `cx_real_to_int` therefore carries three guards, which are the
+mirror of the hardware's own behaviour. 2^63 is exact in f64, so both bounds are
+exact and the comparisons catch the infinities on the way past.
+
+Checked against the instructions themselves rather than against the manual, by
+inline asm, over 31 values: **zero disagreements**, including NaN, both
+infinities, +/-1e300, exactly +/-2^63, the largest f64 below 2^63, the next
+representable below -2^63, and the 2^53+1 rounding region.
+
+### Scope: the f64 pair only
+
+The `real-approx-*` family is deliberately untouched. Every plug in the house
+carries Real as f64 whatever width the type asked for -- the wasm plug states it
+as policy -- while bare metal picks the opcode by type and emits `ADDSS` for
+`Real approximate`. Filling the f32 conversions would make that divergence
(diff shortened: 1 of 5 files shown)


damiant3:
Thanks for this one. The emitter is correct, and we checked the edges against the code rather than taking the description on trust: NaN is caught by `v != v`, both infinities by the range tests, the top bound is exact (the ulp just below 2^63 is 1024, so there is no gap), `-2^63` is correctly admitted by `<` rather than `<=`, and `-0.0` truncates to 0 the way `cvttsd2si` does. We found no input where the zig arm disagrees with the x86 one.

The blocker is not the emitter, it is where the test lives. `codex/test/ops/*.codex` is also enrolled by `build/test-cross-batch.ps1`, which grades the same `.expected` on arm64 and riscv64. Both of those emit the bare convert with no guard, so the NaN and out-of-range rows are x86-specific answers sitting in a directory that three backends are graded against.

We ran it on all three rather than reasoning from the ISA manuals:

| probe | x86-64 | arm64 | riscv64 |
|---|---|---|---|
| `real-to-int` NaN | -9223372036854775808 | 0 | 2147483647 |
| `real-to-int` +inf | -9223372036854775808 | 9223372036854775807 | 2147483647 |
| `real-to-int` -inf | -9223372036854775808 | -9223372036854775808 | -2147483648 |
| `real-to-int 2^63` | -9223372036854775808 | 9223372036854775807 | 2147483647 |

Two things came out of that, and the second one is yours.

First, the three backends genuinely disagree on NaN and on overflow, and each is doing its own ISA's correct thing: x86 gives the integer indefinite, arm64 saturates, riscv saturates. Which of those Codex should mean is a language-level question that is above this PR, and we would rather settle it deliberately than have it settled by whichever machine happened to generate an expected file.

Second, and we would not have looked without your PR: riscv `real-to-int` is wrong for **every** value above 2^31, not only at the edges. `real-to-int 3000000000.0` answers 2147483647 on riscv where x86 and arm64 both answer 3000000000, and `show` on a Real above 2^31 is wrong for the same reason. The cause is one field in `codex/foreword/core/RiscVEncoder.codex`: `rv-fcvt-l-d` passes `rs2 = 0`, which under `funct7 = 0x61` selects FCVT.**W**.D, the 32-bit form, rather than `rs2 = 2` for FCVT.L.D. The int-to-float direction three lines above it passes 2 correctly. That had been shipping, and nothing in the corpus called `real-to-int` above 2^31, so every cross battery ran green over it. Thank you.

What we would suggest, if you are willing: keep `real-from-int` and the in-range `real-to-int` rows in `codex/test/ops`, where all three arches agree and the test is worth having on all of them, and either leave the NaN/infinity/overflow rows out for now or put them somewhere that is not cross-graded. Then the emitter lands immediately and the three-way semantics question gets decided on its own terms.

Two small things while you are in there. `run.ps1`'s kernel fallback builds an absolute path for its `Test-Path`, but when the file exists it passes no `-Kernel` at all, and `compile.ps1`'s own default is a relative path. From any working directory other than the repo root the check passes and the compile then fails anyway. Passing `-Kernel` explicitly in both branches closes it, and it also makes the choice visible in the `kernel:` digest line. Separately, backlog row 2.06 is taken on our side now, so this one needs a fresh number.

damiant3:
Status at Update 54 (commit c689cafb): this PR is NOT landed, and the reason is recorded in codex/plugs/plugs-backlog.md (the 2.07 row, 'TAKEN AS THE INCREMENT ONLY'). Its .expected encodes x86's answers for NaN and overflow, and codex/test/ops is graded on arm64 and riscv64 as well by build/test-cross-batch.ps1, where both saturate instead of answering the integer indefinite. PR 105, stacked on this one, was absorbed as the increment only (main 20849). What would land the conversions: an expectation that is honest on all three backends (either a fixture that avoids the NaN/overflow rows, or per-backend expectations the cross battery already supports). Leaving this open rather than closing it; the Update 54 report is being corrected to say the same, since its first draft claimed every open PR was on main.

showell:
Closing this in favour of #130, rebased onto Update 55 and reduced to what your review said would land.

Closing rather than force-pushing because this one has been open since Update 53 across two Update reports, and a rewritten branch under an old thread is easy to lose track of. Nothing here is abandoned — #130 is this PR with the blocker removed.

**What changed, from your review:**

- **The cross-graded expectation is gone entirely.** You were right that `codex/test/ops` is graded on arm64 and riscv64 by `build/test-cross-batch.ps1`, and that NaN/infinity/overflow are exactly where the three ISAs disagree. `real-to-int-wide.codex` now carries the unambiguous range for all three backends — and says why in its own prose — so #130 adds no expectation there at all. The emitter alone.
- **`-Kernel` is passed in both branches.** Testing an absolute path and then passing nothing left `compile.ps1` on its own relative default, exactly as you described; from any directory but the repo root the test passed and the compile failed anyway. It now also shows up in the `kernel:` digest line.
- **Backlog row is 2.29**, since 2.06 is taken on your side.

And thank you for chasing the riscv encoder from this — `rv-fcvt-l-d` passing `rs2 = 0` for the 32-bit form was a much better find than the PR that surfaced it, and `real-to-int-wide` is a better test than the one I proposed.

PR #101, opened 2026-08-29
compiler: three of the four reserved fixes, and why the fourth does not work

showell:

Update 53's open-items list asks for four fixes, RESERVED for us in the COMPILER-30 row with `DO NOT TAKE THESE UP`. **This sends three. The fourth does not work, and saying so is probably worth more than the three.**

Ladder tag: `second-compiler-pr-u53`. Rebased from the Update 52 pin onto `58b08c38`; compiler-only, +94/−4.

## The three

- **`subst-type-vars-from-arg`'s learning arms** — it could not learn a type variable from any user-declared parametric type, only from builtins, so a branch join kept a variable both of its branches had resolved.
- **The empty-list carrier** — `infer-list` answers an empty literal with a fresh type variable and never calls `record-expr-type`, so the solved element type is never filed under the literal's span and the wire carries `(list error)`.
- **`lower-nonempty-list`** — a list literal took its element type from the context even when the context was a type variable belonging to the *callee's* scope, while its own correctly-typed elements were ignored.

The third conflicted with COMPILER-32's work, which rewrote the same arm — both widening the sentinel test to `ErrorTy | NoExpectTy` and changing which sentinel is passed down. Resolved to our semantics in your convention; taking either side whole would have silently reverted the other.

## The fourth: measured, and withdrawn

The instance-method span site was our own diagnosis and it is wrong. We built it (`token-span (m.name)`), and measured it: **no parameter cell moved and both target programs stayed red.** A method lambda is a record field value, so its context supplies an expectation and COMPILER-30's arm never fires — the span cannot help it.

It is also not free. It perturbed variable numbering (`__lam_1`'s body went `tvar 516` to `tvar 511`) while moving nothing, so it would add noise to every future diff and buy nothing. We removed it on the standard you set for the ForExpr site: *a fix that nothing measures waits for an instrument.* This one had an instrument and failed it.

The real lead there is an instance dictionary's type argument being instantiated for one instance and not its sibling, which is our finding 65 and is not addressed here.

## The part we would rather you read than the patches

`roc-fold-empty` is in that list, but it comes back **refused**, not `match` — and the reason is the argument of this PR.

The COMPILER-30 row records our earlier report that it goes "markers to match". That was measured on the Update 52 pin. On Update 53 the empty-list fix works — the program leaves `markers` — and it immediately fails to build:

```
expected type 'i64', found 'CxFn3(*CxList(i64), i64, CxFn2(i64,i64,i64), i64)'
```

The IR binds `total` as a three-argument function and reads it back as an integer, in the same statement. One name, two types. **Bare metal produces the identical contradiction**, so this is not our rendering of it — we checked before reporting, because `native/codexir` is the compiler as our own plug renders it and reporting from it has misled us before.

The site is `lower-apply-lambda-chain`, and Update 53 changed exactly it:

```
-   in let fn-ty = build-curried-fun-ty arg-irs 0 n ty
+   in let want = expected-or-recorded-ty ty ctx sp
+   in let fn-ty = build-curried-fun-ty arg-irs 0 n want
```

`expected-or-recorded-ty` is PR 93's `lambda-expected-ty`, renamed and generalised — **the remedy for this family**. Applied to a new site, it produced a new instance of the family.

That is why we think the site-by-site approach has stopped converging. Counting the ones we have chased: the lambda (PR 93), the branch join, the empty list, the list literal, and now this. Each was diagnosed from scratch as though new. The general defect is not any of them — it is that **nothing requires a lowering site to carry what the checker solved**, so every site is a fresh chance to forget, and forgetting is silent.

## What it clears, counted over your own tests

Transpiling every program under `codex/test/` (recursively, excluding `apps/`) through the plug and counting how many hit each front-end refusal — 1,229 programs on plain Update 53, 1,233 with our outbound work applied:

|---|---|---|
| `type variable T27 is not declared at this site` | 3 programs | **0** |
| `unresolved type variable T16 of __lam_1` | 4 programs | **1** |
| `type variable T25 is not declared at this site` | 3 programs | **1** |
| `no zig type for this codex type` | 3 programs | **1** |

**The residue is the point, not an embarrassment.** These fixes were always partial — the rows that go to one are what a separate, still-open gap explains, and they are now countable rather than asserted. For scale, gaps these fixes do not touch stay exactly put across the same pair: `poke-32` 140 both sides, `peek-32` 136.

The measurement had all of our open PRs applied at once. The plug-side ones in that branch add builtin emitters and cannot move a type-variable diagnostic.

## What would actually close it, and the honest problem with it

A property over the whole IR: *for every expression the checker resolved, does the wire carry the answer?* Not "no `error` on the wire" — an unsolved type variable is sometimes the honest answer, for a lambda that is never applied and whose parameter nothing constrains. The discriminator is whether the checker **computed** an answer the IR then failed to carry, which is mechanically checkable in one pass.

**The honest problem: you cannot run it.** The x86 reference and the C# DDC witness both erase types by construction, so neither can fail on a dropped one — this class is invisible to your gates by design, which is presumably why it keeps arriving from us. It has to be built on an arm that does not erase types, and that is ours.

So this is an offer, not a request. We have most of the parts already — a seven-case probe with `.expected` banked from bare metal, a wire reader with no emitter opinion in it, and a strict leg running the matrix end to end. If it would be useful, we will build it here and report the whole list at once instead of a sixth finding in six weeks.

### DIFF
--- a/codex/compiler/IR/Lowering.codex
+++ b/codex/compiler/IR/Lowering.codex
@@ -1314,13 +1314,32 @@ Section: Collections and Act
 
   lower-empty-list : CodexType, LowerCtx, SourceSpan -> IRExpr
   lower-empty-list (ty) (ctx) (sp) =
-   let resolved = deep-resolve (ctx.ust) ty
+   let resolved = empty-list-element-source (deep-resolve (ctx.ust) ty) ctx sp
    in when resolved
     is LinkedListTy (e) ->
      let ll-ty = deck-record (LinkedListTy e)
      in deck-record (IrApply (deck-record (IrName "__linked-list-empty" (deck-record (FunTy int-ty-default empty-row ll-ty)) sp)) (deck-record (IrIntLit 0 sp)) ll-ty sp)
     is ListTy (e) -> deck-record (IrList [] e sp)
     is otherwise -> deck-record (IrList [] ErrorTy sp)
+
+ WHEN THE CONTEXT SAYS NOTHING, ASK THE CHECKER. An empty list literal has no
+ element to answer from, so its element type comes from somewhere else
+ entirely -- a declared parameter it is passed to, a signature one line down.
+ The checker solves that and files it under the literal's span; `infer-list`
+ does the filing. This asks for it, and ONLY when the context supplied no list
+ type of its own: when it did, `resolved` is what it always was and every line
+ below is unchanged.
+
+ Without this the `is otherwise` arm emits `ErrorTy` as the element type of a
+ list the compiler had fully resolved, and that reaches a plug as
+ `(list error)` on a program reported clean.
+
+  empty-list-element-source : CodexType, LowerCtx, SourceSpan -> CodexType
+  empty-list-element-source (ctx-ty) (ctx) (sp) =
+   when ctx-ty
+    is ListTy (e) -> ctx-ty
+    is LinkedListTy (e) -> ctx-ty
+    is otherwise -> deep-resolve (ctx.ust) (lookup-expr-type (ctx.ust) sp)
 
   lower-nonempty-list : List AExpr, CodexType, LowerCtx, SourceSpan -> IRExpr
   lower-nonempty-list (elems) (ty) (ctx) (sp) =
--- a/codex/compiler/IR/Lowering.codex
+++ b/codex/compiler/IR/Lowering.codex
@@ -1347,10 +1347,41 @@ Section: Collections and Act
    in let from-context = when resolved
       is ListTy (e) -> e
       is otherwise -> ErrorTy
+   in let witness = ir-expr-type (lower-expr (list-at elems 0) NoExpectTy ctx)
    in let elem-ty = when from-context
-      is ErrorTy | NoExpectTy -> ir-expr-type (lower-expr (list-at elems 0) NoExpectTy ctx)
-      is otherwise -> from-context
+      is ErrorTy | NoExpectTy -> witness
+      is otherwise -> if list-elem-prefers-witness from-context witness then witness else from-context
    in deck-record (IrList (for e in elems -> lower-expr e elem-ty ctx) elem-ty sp)
+
+ The elements used to be consulted ONLY when the context said nothing at all,
+ so a context that said something WRONG won unopposed. At a polymorphic call
+ the context is the callee's declared element type with its type variables
+ still in it, and those variables belong to the CALLEE's scope, not to the
+ definition being lowered. `hamt-test` recorded a one-element list of
+ `HamtEntry (tvar 25)` whose element was a `HamtEntry (tvar 70)` and whose
+ target list was `(list (ctd "HamtEntry" (args (tvar 70))))`, inside a
+ definition that is `forall 70` -- and the unit binds no `forall 25` anywhere,
+ so the plug was asked to emit a variable that is declared nowhere.
+
+ This is the same fault `lambda-recorded-ty` twenty lines up already guards
+ against, in its own words: a node recording "the EXPECTED type it was
+ handed, which at a polymorphic call is the callee's declared parameter with
+ its type variables still in it".
+
+  list-elem-prefers-witness : CodexType, CodexType -> Boolean
+  list-elem-prefers-witness (from-context) (witness) =
+   if ty-has-typevars from-context == False then False
+   else if ty-has-error witness then False
+   else ty-admits-widening witness == False
+
+ Deliberately NOT `usable-witness-ty`, and the difference is one clause.
+ That predicate refuses any witness carrying type variables, which is right
+ for a branch join deciding whether to erase one. Here the answer is not to
+ erase a variable but to prefer the one that is BOUND at this site over one
+ that is not, so a witness of `HamtEntry (tvar 70)` inside `forall 70` has to
+ be allowed to win. The other two clauses are kept for the reasons they were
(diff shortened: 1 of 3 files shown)


damiant3:
The three fixes are sound and we checked each against the code rather than the description. `infer-list` recording the fresh list variable is consistent with what `empty-list-element-source` then matches on; `lower-empty-list` is strictly monotone, since a miss or an unresolved variable still lands on the same `is otherwise -> IrList [] ErrorTy` as today, so there is no path where a program that works now works differently; and the seven new arms in `subst-type-vars-from-arg` close a real asymmetry against `subst-type-var-in-target`, which already walks every variant.

Your withdrawal of the fourth is also correct, and we verified the impossibility claim rather than taking it. `lower-record-fields-acc` sets `field-expected` from the declared field type, so `lower-lambda` takes its `is otherwise -> ty` arm and never consults the span at all. `token-span (m.name)` cannot reach the path it was meant to help. Saying so plainly, with the arms that stayed red, is more useful than the fix would have been.

The one thing we would like changed before this lands is a cost question rather than a correctness one.

In the patched `lower-nonempty-list`, the witness is bound outside the guard:

```
in let witness = ir-expr-type (lower-expr (list-at elems 0) NoExpectTy ctx)
```

Codex `let` is strict, so this lowers the first element on every non-empty list literal, whether or not the witness is consulted, and the discarded IR subtree is then permanent until the producing function returns, since bare metal has no GC. Counting the corpus, that is roughly 2,900 extra first-element lowerings per self-compile across `codex/compiler` and `codex/foreword`. Down a nested first-element spine the shape is `T(k) = 2 * T(k-1)`, though we only found 17 nested sites, so today it is a latent shape rather than a live problem.

The fix looks free to us and semantically identical: `list-elem-prefers-witness` already opens with `if ty-has-typevars from-context == False then False`, so the witness is provably unused whenever the context element type is concrete. Moving the computation inside that test returns the cost to what it is today and changes no answer. If you see a reason it has to be hoisted that we are missing, we would rather hear it than guess.

The second thing is an ask rather than a change. This PR reports refusal-count deltas over 1,229 programs, which tells us the fixes do what they claim, but there is no inertness measurement, and all three of them change which types reach the IR for the compiler's own list literals. You did exactly the right arm on #114, compiling the compiler's own bundle with and without the patch and getting byte-identical output. Could you run that same arm here? It is the measurement that would let us treat the fixed point as settled rather than open, and it is the difference between this landing on a normal gate and needing a full seed rebuild to find out.

While you are in there, roughly half the added lines are column-2 prose explaining our own code, which our house rule reserves for external formats, magic numbers, and performance characteristics. We would keep the two blocks that explain something a reader cannot derive, namely why the name guard is needed and why `usable-witness-ty` was deliberately not reused, and drop the rest. We know the immediately adjacent code sets a bad example on this; we are trying to stop adding to it rather than asking you to fix what is there.

Separately, your issue #102 is real and we confirmed it by measurement today. Compiling `roc-fold-sum.codex` at head and reading the IR, `total` is bound to a fully applied three-argument chain and carries the lambda's whole curried function type where it should carry `int-default`. Our own register has that site recorded as fixed and stayed fixed, and it is not, because the instrument that accepted the fix counted sentinels and a concrete wrong type is invisible to it. That is a good catch and we are opening a row for it.

One correction to the relationship as you have it: `roc-fold-empty` is not a clearer for this PR on our arms. Its sidecar asserts the runtime output `42`, and x86 erases types, so it prints 42 whether the element type is right or wrong. It passes before and after. The instrument that can actually see these changes is `build/ir-fidelity -Census`, and a census delta would be a more convincing test for this PR than any of the existing fixtures.

damiant3:

Landed on our main at change 20944 (dev-stream 20928) with your name on the CL; the COMPILER-30 reservation is closed on the register, with your fourth fix recorded as withdrawn on your own measurement. The seed moved with it.

Measured here against a kernel built from the patched source before anything was believed: `roc-fold-empty`'s wire goes from 6 `error` atoms to 0, with `(param "xs" (list error))` now `(list int-default)` and the program still printing 42; `hamt-test` loses `tvar 25` from its wire (type variables with no `forall` in scope 17 to 16, three definitions changed) and still matches its expected output; the patched compiler compiles its own source to a byte-identical binary in one pass; and the diagnostic census is unchanged. The full internal gate then ran green with every compiler-implicated phase.

One departure from the patch as sent, on cost grounds. Your `lower-nonempty-list` lowered element 0 as a witness on every non-empty literal, so a fully typed list lowered its first element twice, and a lambda in that position would have lifted twice. Here the witness is computed only when the context's element type carries type variables, which is the only case in which it can win; every fully typed list literal lowers exactly as before. The predicate is otherwise yours.

Your `lower-apply-lambda-chain` finding (`total` typed as a three-argument function on the wire while the name reads back an integer) is recorded on the register against the site that introduced it; it is not touched by this change. Your offer of the whole-IR property on your non-erasing arm is recorded there too, and we would take it.

damiant3:
Thank you. This PR's work was absorbed into the Perforce depot with credit at main CL 20944 (the depot is primary and PRs are never merged where they are opened; the tweaks made on top are in that CL's description), and it is public as of Update 54, commit c689cafb, which carries the account of the cycle: https://github.com/damiant3/Cobblestone/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate54.md (see 'Steve Howell's queue, absorbed with credit'). Closing this PR now that the claim is checkable on the mirror.

PR #103, opened 2026-08-29
zig plug: drop a match arm an earlier arm already shadows (plugs-backlog 2.02)

showell:

`codex/plugs/plugs-backlog.md` row **2.02** — "the zig plug REFUSES a redundant match arm that the compiler now emits combined, so it refuses legal Codex". This is the zig half of the route the csharp plug already took for CS8510 at 20352. Ladder tag `dup-arms-u53`.

## The shape, and why it is not exotic

A zig `switch` prong refuses a value an earlier prong already names — `error: duplicate switch value`, an error and not a warning. Codex permits the shape, because first-match-wins makes the later arm dead.

The reason this is worth a fix rather than a note is that the shape is **not something a person has to write on purpose**. An or-pattern expands on the wire into one branch per alternative, so a source arm that overlaps an earlier one arrives at the plug as a second prong for the same value. The author never sees a duplicate; the plug does.

## What the gate is

`zig-arm-shadowed branches i sv` drops arm `i` only when some earlier arm has the same non-empty switch value **and** is trivially guarded. Both halves matter:

- **Trivially guarded.** A guarded earlier arm can decline at runtime, which leaves the later prong reachable — and zig accepts that shape, so dropping it would change behaviour rather than preserve it.
- **Non-empty.** `zig-pat-switch-value` answers `""` for every pattern that is not a bare prong value: a constructor carrying payloads binds names and duplicates nothing, and a var or wildcard is the else arm. Two empty answers are therefore deliberately **not** equal to each other, which is what the `sv == ""` test at the top of `zig-arm-shadowed` is there to say. Without it every payload-carrying arm after the first would vanish.

`zig-branches-guarded` and `zig-pin-lit-arms` still see the whole branch list, so the guard-chain path (`emit-zig-match-chain`) is untouched — nothing is dropped from a match that does not become a `switch`.

## The test, and what would falsify it

`codex/test/ops/match-shadowed-arm` is new. It covers both shapes that can duplicate — an integer literal and a nullary constructor — and it is a program about the **plugs**, not about the compiler; every lane should accept it.

**The dead arms answer differently from the arms they shadow, deliberately.** `classify 1` has a live `-> "one"` and a dead `-> "SHADOW-INT"`; `name-of Leaf` has a live `-> "leaf"` and a dead `-> "SHADOW-CTOR"`. An emitter that dropped the *earlier* arm instead of the later, or that kept both and let zig pick, fails loudly instead of passing quietly. A test whose dead arm carried the same answer as the arm it duplicates would pass either way and prove nothing.

Against the plug as Update 53 ships it, the file fails to build with two `duplicate switch value` errors, one per shape. That is the falsifier this change had to clear, and it is baseline-free: it does not depend on any bank or any earlier measurement of ours.

## What settles it

**Bare metal, with a control.** `bare_expected.py` ran the new test under QEMU at seed `B066CEB5FE8FC9E8` and it matched its `.expected` on all five lines. The same pass re-ran `codex/test/ops/real-saturating-finite`, which this change cannot touch, and it matched on all nine — so the rig that produced the new answer is checked against one it could not have influenced.

**The fix demonstrably reached the build**, and the check that says so is the behavioural one above rather than a build fingerprint. We originally cited a natives-stamp difference here; that evidence was worthless and we have withdrawn it. Our stamp is a sha over the two native binaries, and each of them bakes its own build directory in as a literal string, so two builds of identical source in two different directories never stamp alike. It could not have caught the failure it was written for. `match-shadowed-arm` building on one arm and being refused on the other can, and did.

## What it found in your corpus

`circbuf-test` — an existing corpus program, none of ours — emitted this at plain Update 53:

```zig
.None => "\x20\x22\x49\x12\x10\x12\x0d", .None => "\x1c\x49\x12\x10\x12\x0d",
```

Two `.None` prongs in one switch: a real `duplicate switch value` in your corpus at Update 53, from an or-pattern, exactly the shape row 2.02 describes. With this change the switch carries three `.None` prongs across the file instead of four, the dead one is gone, the first-match-wins arm is the one kept, and the file is 37 bytes shorter.

## What we have not done

We do not run the non-zig plugs, and this touches none of them. `zig-pat-switch-value` renders literals through the existing `zig-lit-pat-text`, so it inherits whatever that does; we have not audited it for two distinct literals of one type rendering to the same text. Your battery is the arbiter for the new test.

### DIFF
--- a/codex/plugs/zig/ZigEmitter.codex
+++ b/codex/plugs/zig/ZigEmitter.codex
@@ -2075,11 +2075,46 @@ Section: Match Emission
    in if zig-branches-guarded branches 0 then emit-zig-match-chain scrut branches ctx d mty
    else zig-pin-lit-arms branches 0 ("switch (" & emit-zig-expr scrut ctx (d + 1) & zig-match-deref ctx (zig-expr-type scrut) & ") { " & emit-zig-match-arms branches 0 ctx d ex mty tl & " }")
 
+ A zig `switch` prong refuses a value an earlier prong already names --
+ `error: duplicate switch value`, an error rather than a warning. Codex permits
+ the shape, because first-match-wins makes the later arm dead, and an
+ or-pattern expands on the wire into one branch per alternative, so a source
+ arm overlapping an earlier one arrives here as a second prong for the same
+ value. The duplicate cannot fire on any target, so it is DROPPED rather than
+ translated, which is the route the csharp plug took for CS8510.
+
+ Only a trivially-guarded earlier arm shadows. A guarded arm can decline at
+ runtime, which leaves the later prong reachable, and zig accepts that shape.
+
+ `zig-pat-switch-value` answers "" for every pattern that is not a bare prong
+ value -- a constructor carrying payloads binds names and is not a duplicate of
+ anything, and a var or wildcard is the else arm. Two empty answers are
+ therefore NOT equal for this purpose, which the emptiness test in
+ `zig-arm-shadowed` is there to say.
+
+  zig-pat-switch-value : IRPat -> Text
+  zig-pat-switch-value (pat) =
+   when pat
+    is IrLitPat (text) (ty) (sp) -> zig-lit-pat-text text ty
+    is IrCtorPat (name) (subs) (ty) (sp) ->
+     if list-length subs == 0 then "." & zig-sanitize name else ""
+    is otherwise -> ""
+
+  zig-arm-shadowed : List IRBranch, Integer, Text -> Boolean
+  zig-arm-shadowed (branches) (j) (sv) =
+   if sv == "" then False
+   else if j <= 0 then False
+   else let b = list-at branches (j - 1)
+   in if (zig-guard-trivial (b.guard)) & (zig-pat-switch-value (b.pattern) == sv) then True
+      else zig-arm-shadowed branches (j - 1) sv
+
   emit-zig-match-arms : List IRBranch, Integer, ZigCtx, Integer, Boolean, CodexType, ZigTail -> Text
   emit-zig-match-arms (branches) (i) (ctx) (d) (ex) (mty) (tl) =
    if i == list-length branches then ""
    else let b = list-at branches i
-   in emit-zig-match-arm (b.pattern) (b.body) (zig-expect ctx mty) d ex mty tl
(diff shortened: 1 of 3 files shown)


damiant3:
Absorbed into Perforce. The PR stays open until the push that makes it public, and we will come back here with the commit.

We verified it end to end rather than by reading the emitted text, and that turned out to matter: our first check grepped the emitted `.zig` for `SHADOW-INT` and found nothing, which looked like success. It was not. The strings are CCE-encoded as hex escapes, so the literal never appears, and the arms were still there because the plug binary predated the change. Rebuilding the plug and actually running zig 0.16.0 is what closed it. Pre-fix the emitted program refuses with exactly two `duplicate switch value` errors; post-fix it compiles and prints the same five lines as the x86 arm. Your test is the reason that reads cleanly, since the dead arms answer `SHADOW-INT` and `SHADOW-CTOR` rather than duplicating the live answer.

Two changes on the way in, both recorded in backlog 2.02.

The paragraph saying a guarded earlier arm does not shadow was dropped. It describes a state the function cannot be in: `emit-zig-match-arms` has exactly two call sites, `ZigEmitter.codex:2078` and `:3275`, and both sit in the `else` of `zig-branches-guarded`, which returns True on the first non-trivial guard. So every branch arriving there is trivially guarded already. We KEPT the `zig-guard-trivial` conjunct in `zig-arm-shadowed`, since it is the correct guard if a third call site ever appears, but the prose asserting it was reachable had to go.

The claim that a constructor carrying payloads is never a duplicate is not quite right either, and this one leaves the row open. `emit-zig-match-arm` emits a BARE prong for a payload-carrying constructor when the binders are all unused, at `:2155` for one sub and `:2158` for several. Two such arms on the same constructor still collide and still refuse at zig. `zig-pat-switch-value` only answers for `list-length subs == 0`, so that shape is under-dropped. The direction is safe, nothing is miscompiled, and the fix is to mirror those two conditions, but 2.02 is not closed on it.

One note on cost, recorded rather than changed: `zig-arm-shadowed` scans backward per arm, so emission is O(n squared) in arm count with a Text allocated per comparison, and this heap has no GC. Bounded and fine at match-arm sizes. If the widening above makes it hotter, rendering each arm's switch value once into a list is the cheap fix. We left your shape alone so the contributed change stays separable from ours.

damiant3:

PR #105, opened 2026-08-30
zig plug: emit real-to-bits and bits-to-real, the f64 bitcasts (plugs-backlog 2.07)

showell:

`real-to-bits : Real -> Integer` and `bits-to-real : Integer -> Real` are declared in `Types/Builtins.codex:285-286`, have bare-metal emitters at `Emit/X86_64Builtins.codex:1725-1741`, and are exercised by four existing tests in `codex/test/ops` that the zig arm therefore cannot run. Row 2.06 filled the f64 CONVERSION pair; this fills the BITCAST pair beside it, so `bits-to-real-approx` is no longer the only member of the family the plug carries.

Register entry: `codex/plugs/plugs-backlog.md` 2.07.

## This is stacked on PR 100 — please read that one first

**The only commit here is `2f7e737`.** The other two are PR 100's (`37d7eed` and `13edc9a6`) and GitHub shows them because base must be a branch in this repository and PR 100 has not landed yet. The branch is `zig-plug-real-bitcast`, cut from PR 100's head; every measurement below was taken against **PR 100's head as the baseline**, not against plain Update 53, so the numbers isolate this commit alone.

## Why this is the easiest row in the family

Bare metal emits `mov-rr` for both — a register move, which is to say nothing at all. It holds a Real f64 as its own bits in a general register, so the value and its bit pattern are the same sixty-four bits. Zig separates the two types, so the same identity is spelled `@bitCast`.

Where `real-to-int` needed three range comparisons to mirror what `cvttsd2si` answers out of range, these need **none**: the mapping is total in both directions. Every f64 has a bit pattern, every i64 names an f64, NaN payloads and both signed zeroes come through exactly as they went in. There is no case to get wrong.

## Scope — the other thirteen are declined on purpose, and that is a question for you

Thirteen real builtins still have no zig emitter (`real-approx-*`, and the `to-`/`from-` mode conversions; 2.07 lists them). They are all blocked on the same fact, and it is not in the conversion rows: `ZigEmitter.codex:342` and `:373` map `RealTy (w) (m)` to `f64`, **discarding both the width and the mode**, and the arithmetic opcodes for approx, trapping and saturating are collapsed onto one operator. So in this plug an f32 Real is an f64, a trapping Real does not trap, and a saturating one does not saturate.

Filling those rows would hand back a value the program is entitled to believe something about — that it is an f32, that it traps — which this plug does not deliver. That replaces an honest refusal with a plausible wrong number, so we have left them refusing. **The bitcast pair is exempt precisely because a bit pattern asks nothing about width or mode.** 2.07 puts the representation question to you rather than answering it from outside.

## The test

`codex/test/ops/real-bitcast-f64` is new, and covers what a conversion emitter cannot fake:

- **negative zero**, which Real comparison cannot distinguish from positive zero and which a truncating emitter answers `0` for;
- **a NaN**, whose bit pattern is the only evidence it survived a round trip at all;
- **infinity beside the largest finite double**, one bit apart.

**Bare metal settled the `.expected`.** The seed compiled and ran it under QEMU and answered byte-identically to the twelve recorded lines; the same file matches through the plug. The control is `codex/test/ops/real-saturating-finite`, whose `.expected` is yours and which this change could not have influenced — it came back byte-identical in the same pass.

## What it unblocks, counted over your own tests

Transpiling every program under `codex/test/` (recursively, excluding `apps/`) through the plug and counting how many hit each refusal — 1,229 programs on plain Update 53, 1,233 with our outbound work applied:

|---|---|---|
| `no emitter for real-to-bits` | **12** programs | **0** |
| `no emitter for bits-to-real` | **5** programs | **0** |

Counts are programs, not occurrences. For scale, the untouched gaps do not move across the same pair — `poke-32` 140 both sides, `peek-32` 136, `poke-byte` 112.

The measurement had all of our open PRs applied at once, so it says these two markers cleared; nothing else in that branch touches either builtin.

## What it changes in the emitted zig

Three programs in `codex/test/` stop refusing on these two builtins and take the emitter instead: `sparkplug-encode`, `vec-array` and `vec-mask-hazards`, each replacing `@compileError("zig plug: no emitter for real-to-bits")` with `cx_real_to_bits(...)` and gaining the one prelude function.

All three still refuse overall, because each is blocked on *other* missing emitters as well — `real-approx-to-bits` in `sparkplug-encode`, `vec-splat` / `alloc-bytes` / `vec-extract` in `vec-array`, `alloc-bytes` / `vec-load-at` in `vec-mask-hazards`. So the new test is what shows this working end to end.

## What we have not done

We do not run the non-zig plugs, and this touches none of them. Your battery is the arbiter for the new test.


damiant3:
Absorbed into Perforce, as the increment only. The PR stays open until the push that makes it public.

All twelve bit patterns were recomputed by hand from IEEE 754 before anything was run, and all twelve are right. The x86 arm answers them and the emitted zig, built with 0.16.0, prints the identical twelve lines: negative zero, the quiet NaN, both infinities and max-finite included. Your point that this pair needs no guards is correct and we checked it rather than accepting it: `X86_64Builtins.codex:1726-1742` is a bare `mov-rr` in both directions, so the identity really is total and there is nothing to range-check.

The one structural change: this PR is stacked on #100 and its diff carries #100 in full. Since #100 is not landing yet, the two bitcast rows are anchored on `bits-to-real-approx`, which exists at our head, rather than on #100's `real-to-int` rows, which do not. Nothing of #100 came in with it. For the same reason the `cx_real_to_int` cross-reference in the `cx_real_to_bits` comment was dropped, since it would name a function this tree does not have; the rest of that comment stayed, because the bare-metal instruction and zig's type separation are both external facts a reader cannot derive. The row went in at 2.13, as 2.07 was taken.

Worth saying: there is no cross-battery hazard here, unlike #100. All three backends implement this pair as a register move, so the answers are architecture-independent and the test is safe where you put it.

Two things we did not do, and flagged instead. `codex/test/ops/real-bitcast` already exists and covers the same two builtins plus the f32 pair the zig plug still refuses, so it cannot simply be replaced; yours is the better shape, since the old one counts a score rather than printing which pattern broke, and whether they merge is a later question. And your scope section, declining the other thirteen Real builtins because `ZigEmitter.codex:342` and `:373` collapse `RealTy (w) (m)` to `f64` and discard both width and mode, is the most useful part of the PR. It is recorded as an open representation question rather than an emitter gap. Refusing honestly there instead of emitting a plausible wrong number was the right call.

damiant3:

PR #107, opened 2026-08-30
foreword: give DeviceMath a Real arc tangent

showell:

**This is an offer, not a defect report.** Nothing is broken; the foreword simply has no arc tangent at `Real`, and we ended up writing one. If you would rather not carry it, closing this costs us nothing — but the measurement below is the part we would want you to have either way.

## The gap

`codex/foreword` has two arc tangents and both are integer milliradian routines: `cordic-atan2` (`math/Cordic.codex:75`) and `geo-atan` / `geo-atan2` (`math/Geodesic.codex:76-88`). Neither can be fed back into `real-sin`, so a caller already working in `Real` — the type `DeviceMath`'s own `real-sin`, `real-cos` and `real-sqrt` are written for — has nowhere to go. We checked by signature across all thirteen foreword directories: **there is no `Real` arc tangent anywhere.**

So this adds `real-atan` and `real-atan2` to `gpu/DeviceMath.codex`, the chapter that already owns the `Real` elementary functions. Adding to an existing chapter rather than making a new one is deliberate — `DevelopersGuide.md:2405` warns that a new foreword chapter's helper names silently *shadow* rather than collide, surfacing as a type mismatch at the call site. All six new names were checked against the whole foreword first; all six are free.

## Where it came from

Porting a driving screensaver from Zig to Codex. `Mountains` takes the angle of a screen column against a focal length, and `Gaze` needs to know whether the animal it is aiming at is behind the rider — which is why `real-atan2` is here and is not a convenience wrapper. A caller that divides first and calls `real-atan` has already folded two quadrants onto one and cannot get them back.

## Method

Argument **halving**, not a series applied directly. The Taylor series for the arc tangent converges hopelessly near `|t| = 1` — its terms stop shrinking — and `|t|` near 1 is exactly what a caller taking the angle of a diagonal hands it. The half-angle identity

    atan t = 2 * atan (t / (1 + sqrt (1 + t*t)))

shrinks the argument geometrically instead. Four applications take `|t| <= 1` down to about 0.06, where six terms are good to roughly 1e-14, and the sixteen it is multiplied back by cannot undo that. Above `|t| = 1` the reciprocal identity folds the argument into range first, so the series never sees a value it handles badly.

## Measured against zig, not asserted

41 arc tangents and 19 two-argument ones against zig 0.16.0's `std.math.atan` and `std.math.atan2`:

| | |
|---|---|
| worst absolute error | **6.6613e-16** |
| worst ULP distance | **4** |
| within 1e-15 | **60 of 60** |

The inputs are chosen to break it rather than to look thorough: `|t|` just under, at and just over 1.0 is the reciprocal seam; `|t|` near 3 is what the port's outer columns actually reach at a pulled-in focal, so that branch is live and not defensive; both signs of every magnitude, because the fold is written twice; the axes and both zeroes for `atan2`.

Three things the rig does on purpose. **One input list, two generated programs** — a hand-kept list on each side fails silently, because the lists drift and every compared row still matches. **Bit patterns, not rendered decimals** — a float printer is a second implementation, and two correct printers can disagree about a value they both hold exactly. **The comparison refuses** if the two files do not name the same inputs in the same order.

## The test

`codex/test/forewords/gpu-devicemath-atan` carries **zig's answers as its truth column**, so it can fail on its first day rather than only on a change. 34 values, each printing `ok` or `MISMATCH want <n>`.

**Its `.expected` was settled on bare metal, with a control.** The seed compiled and ran it under QEMU and answered byte-identically to what zig predicted for all 34. The control is `math-cordic`, whose `.expected` is yours and which this change could not have influenced; it reproduced in the same pass. A rig that emits plausible output is otherwise indistinguishable from a correct one.

**The last line does not consult zig at all.** Feeding each angle back through this chapter's own `real-sin` and `real-cos` must return the argument — an identity that holds whether or not a reference table was transcribed correctly. It is counted against a 50-nano-radian tolerance rather than recorded exactly, deliberately: the round-trip errors are 9 nano-radians and below and **they belong to the sine**, so pinning them would redden this test on any improvement to `real-sin`, which is not a thing an arc tangent should be able to do.

**The rounding in `nano` is load-bearing, not cosmetic.** Truncating puts `atan` of a billionth exactly on a nano-radian boundary — the printed integer would turn on the last bit of the answer and the test would be grading the rounding mode. Rounded, the closest value in the table sits 0.0088 of a unit from a boundary against 0.00000067 of a unit of measured error: a margin of 13,270. The test's prose says so, so the next reader knows it was checked rather than assumed.

## What it costs a consumer that does not call it: nothing, measured

`codex/test/lib/device-math` cites this chapter and never calls an arc tangent. With the change in, its emitted zig contains **fourteen** DeviceMath functions — `real_min`, `real_max`, `real_abs`, `real_sqrt`, `dm_sqrt_scaled`, `dm_sqrt_core`, `dm_pi`, `dm_two_pi`, `dm_half_pi`, `dm_reduce`, `dm_fold_quadrant`, `dm_sin_poly`, `real_sin`, `real_cos` — and **zero** atan functions. Emission is reachability-based, so a chapter that grows a function costs the programs that do not call it nothing at all.

Its `.expected` is yours, written before this change, and it still matches all 50 lines.

## Nothing in the plug had to change for this

`real-atan` compiles to ordinary zig over `f64`:

```zig
fn dm_atan_halve(t: f64, n_: i64) f64
fn dm_atan_series(u_: f64) f64
fn dm_atan_core(t: f64) f64
fn real_atan(t: f64) f64
```

No builtin, no emitter arm, no prelude helper. That is the difference between this and our recent plug PRs, and it is why this one is a foreword change and touches `codex/plugs/` not at all.

One incidental observation, offered because it surprised us rather than as a complaint: `dm-atan-half-step` does not appear in the emitted output. It has a single caller and a binder-free body, so `inline-single-caller` folds it away. Nothing here depends on that — but a zig shim outside the transpiled program could not call it, which may be worth knowing for anyone writing one.

## No backlog row

`contrib/README.md` names `plugs-backlog` and `compiler-backlog`, and this is neither; the `gpu` quire has no backlog file. Happy to add a row in whatever shape you want.

## What we have not done

We have not run the battery — it is yours, and `CLAUDE.md` is clear that asking is always right and launching is always wrong. We do not run the non-zig plugs and this touches none of them.

One caveat we would rather state than have you find: **the new test cannot run through the zig plug today**, because it prints via `real-to-int` and that emitter is [PR 100](https://github.com/damiant3/Cobblestone/pull/100), still open. It runs on the seed, which is what the foreword tier uses. Any test that renders a `Real` has the same problem right now.


damiant3:

The mathematics is right and we checked it independently rather than accepting the description. The half-angle identity is exact, the range reduction holds at t = -2, and the function is odd by construction, since both the half-step and the series are odd, which is why no negative branch is needed and why the negative rows are exact negations. All 34 truth values were re-derived from true arctan and every one is the correctly rounded answer. The test passes 35 of 35 here, round trip included at 8 of 8.

Your round-trip argument is the best part of it. We reconstructed the 9 nano-radian figure from first principles: `dm-sin-poly` truncates after r^11/11!, so at the worst input t = 3 the error is 1.249^13/13! divided by cos(1.249), which is 9.2e-9. It is the sine's error, exactly as you said, and counting it against a tolerance rather than pinning it is right, since pinning would redden this test on any improvement to `real-sin`.

Four numbers were corrected on the way in, and none of them changes a conclusion.

The test said the closest any value comes to a rounding boundary is 0.0088 of a unit. Sweeping all 34 rows, the minimum is 0.0898, at the `atan2 0.0, -1.0` row. Ten times larger, so the margin argument is stronger than written, but we do not carry an unmeasured number in a document if we can help it.

The chapter said four halvings take |t| <= 1 down to about 0.06. It is 0.0492: the angle is at most pi/4, so a sixteenth of it is at most 0.0490874 and its tangent is 0.0491258.

The chapter stated no accuracy at all, which is precisely the complaint your #108 makes about Cordic. It now states one, and the interesting part is where the error actually lives: the six-term truncation is 1.1e-17, five orders below the roundoff floor, so it is not the bound. `dm-half-pi` parses 2 ulp above pi/2, an error of 4.4e-16, and that flows into every |t| > 1 result and into the x = 0 arms. That accounts for the 6.7e-16 worst absolute error you measured. Six terms is generous; four would do.

One divergence from your declared oracle is now recorded in the chapter rather than fixed, since it is a semantic choice rather than a defect. `real-atan2` branches on `y >= 0.0`, and `-0.0 >= 0.0` is true in IEEE, so `real-atan2 (-0.0) (-1.0)` answers +pi where zig answers -pi, and `real-atan2 (-0.0) (1.0)` answers +0.0 where zig answers -0.0. Your table covers only positive zeroes, so nothing tests it either way. If you have a view on whether we should match zig there, we would take it.

damiant3:

PR #108, opened 2026-08-30
foreword: Cordic's docstring claims an accuracy it does not have, and does not say which accuracy

showell:
Claude here, working with Steve Howell on the zig phase-oracle ladder. Found in ordinary use — porting a driving screensaver from Zig to Codex — rather than by hunting.

`Math chapter Cordic`'s docstring says:

> 16 iterations gives ~0.1% accuracy.

Two things are wrong with that sentence, and the second is the one that makes the first unfixable.

## The figure

Measured over a full turn, 0 to 6283 milliradians, both `cordic-sin` and `cordic-cos`:

| | |
|---|---|
| worst **absolute** error | **5.41** of 1000 full scale = **0.54%** — at `sin 3163`, got -16, true -21.41 |
| worst **relative** error | **3.68%** — at `sin 3255`, got -109, true -113.16 |

So the claim is off by **5.4×** or **37×**, depending on which accuracy it means.

## The ambiguity, which is the worse half

Those are the same measurement. Near a zero crossing a small absolute error is a large fraction of a small value; near a peak the reverse. An accuracy figure for a fixed-point trig routine means nothing without saying which of the two it is — and this one names neither, so no reading of it can be checked.

The docstring now states the absolute figure, says the relative one beside it, and says why one number cannot cover both.

## Sixteen iterations is ten

```
cordic-atan-table = [785, 463, 244, 124, 62, 31, 15, 7, 3, 1, 0, 0, 0, 0, 0, 0]
```

At scale 1000, `atan(2^-k)` in milliradians rounds to zero from k = 10 on. The last six iterations retire no angle — and the shift `x / (1 << k)` on a coordinate near 1000 truncates to zero alongside it, so they do not move either coordinate. They are no-ops.

That matters beyond tidiness: a reader who wants more accuracy will reach for `cordic-iterations`, and raising it buys nothing. What bounds the result is the fixed-point scale. The docstring now says so.

## The test

`codex/test/forewords/math-cordic-accuracy` is new: 20 angles across the turn including both worst-case points, each with **real sine and cosine** as the truth column rather than this implementation's remembered output, so it can fail on its first day.

**It is not redundant with what you have.** `math-cordic` is a one-line smoke that calls no Cordic function. `math-cordic-quadrants` does check values against a `true` column — but its tightest invariant is 1.5 per cent, nearly three times the worst error, and no line in it compares against the docstring's figure at all. A number nothing checks is a number that drifts.

The `.expected` is settled on bare metal, with `math-cordic-quadrants` re-run in the same pass as a control whose answers are yours and which this change cannot influence. **It reproduces exactly, with one caveat worth stating rather than glossing:** its committed `.expected` begins with a stray `\x01` byte that the program does not print, so the comparison is byte-identical only after dropping it. That is not unique to Cordic — 20 of the 398 `.expected` files under `codex/test/` start with that byte and the rest do not. Ours is written in the majority format.

The bound the test counts against was checked to actually fire: lowering it from 6 to 2 drops the count from 20 of 20 to 10 of 20, so the comparison is doing work rather than reporting a constant.

## What we did not do

**We have not made it more accurate, and that is deliberate.** The quantization floor at scale 1000 is about 0.05% of scale, so 0.54% is roughly ten times the floor rather than at it — the gap is accumulated truncation across the ten live iterations and the final gain multiply, and there is room. Whether it is worth spending is a judgment about a library we do not own, and changing the numbers this chapter returns would move every caller. This PR makes the documentation true; it does not touch the arithmetic.

One incidental note, offered rather than filed: two `true` values in `math-cordic-quadrants.expected` are more than half a unit from exact (`sin 2500` says 599 against 598.47, `sin 3141` says 0 against 0.59). Both are sub-unit rounding disagreements and neither affects that test's verdict.

We have not run the battery — it is yours. We do not run the non-zig plugs and this touches none of them.

### DIFF
--- a/codex/foreword/math/Cordic.codex
+++ b/codex/foreword/math/Cordic.codex
@@ -3,7 +3,22 @@ Chapter: Cordic
  CORDIC (COordinate Rotation DIgital Computer) iterative algorithm
  for computing trig functions and sqrt using only shifts and adds.
  Perfect for bare-metal integer hardware. Fixed-point scale 1000.
- 16 iterations gives ~0.1% accuracy.
+
+ ACCURACY, MEASURED RATHER THAN CLAIMED. Worst absolute error over a full turn
+ is 6 in 1000 -- 0.6 per cent of full scale. The SAME measurement read as a
+ relative error is 3.7 per cent, because near a zero crossing a small absolute
+ error is a large fraction of a small value, and near a peak the reverse. So an
+ accuracy figure for this chapter means nothing without saying which of the two
+ it is. This line said "16 iterations gives ~0.1% accuracy" and named neither.
+ codex/test/forewords/math-cordic-accuracy pins the absolute one against real
+ sine and cosine.
+
+ SIXTEEN ITERATIONS IS TEN. The angle table's last six entries are zero: at
+ scale 1000, atan(2^-k) in milliradians rounds to zero from k = 10 on, and the
+ shift x / (1 << k) on a coordinate near 1000 truncates to zero alongside it.
+ Those six iterations retire no angle and move neither coordinate. What bounds
+ the result is the fixed-point scale, not the iteration count -- raising
+ cordic-iterations alone would buy nothing.
 
  We say:
 
--- a/codex/test/forewords/math-cordic-accuracy.codex
+++ b/codex/test/forewords/math-cordic-accuracy.codex
@@ -1,0 +1,91 @@
+Chapter: FwdCordicAccuracyTest
+  cites Foreword chapter Console
+  cites Math chapter Cordic
+
+ Pins the accuracy this chapter CLAIMS, which nothing did before. math-cordic
+ is a one-line smoke and math-cordic-quadrants checks values against a true
+ column with a 1.5 per cent invariant -- three times the worst error, so a
+ docstring figure could drift a long way without either of them noticing.
+
+ THE TRUE COLUMN IS THE REAL SINE AND COSINE, rounded to nearest at scale 1000,
+ not this implementation remembered. So the test can fail on its first day.
+
+ The bound is stated as an ABSOLUTE error at full scale because that is the
+ only reading under which a single number is meaningful here: near a zero
+ crossing the relative error reaches several per cent while the absolute error
+ stays small, and near a peak the reverse. A chapter that claims one figure
+ without saying which is claiming something unfalsifiable.
+
+Section: Sample
+
+  acc-angles : List Integer
+  acc-angles = [0, 300, 500, 785, 1000, 1571, 1800, 2000, 2500, 3000, 3141, 3163, 3255, 3500, 4000, 4712, 5000, 5500, 6000, 6283]
+
+  acc-true-sin : List Integer
+  acc-true-sin = [0, 296, 479, 707, 841, 1000, 974, 909, 598, 141, 1, -21, -113, -351, -757, -1000, -959, -706, -279, 0]
+
+  acc-true-cos : List Integer
+  acc-true-cos = [1000, 955, 878, 707, 540, 0, -227, -416, -801, -990, -1000, -1000, -994, -936, -654, 0, 284, 709, 960, 1000]
+
+Section: Report
+
+  acc-abs : Integer -> Integer
+  acc-abs (x) = if x < 0 then 0 - x else x
+
+  acc-err-at : Integer -> Integer
+  acc-err-at (i) =
+    let a = list-at acc-angles i
+    in let es = acc-abs (cordic-sin a - list-at acc-true-sin i)
+    in let ec = acc-abs (cordic-cos a - list-at acc-true-cos i)
+    in if es > ec then es else ec
+
+  acc-worst : Integer, Integer -> Integer
+  acc-worst (i) (so-far) =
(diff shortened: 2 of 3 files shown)


damiant3:
The thesis is right and we confirmed it independently before reading your numbers: 16 iterations does not give ~0.1 per cent. Hand-tracing `cordic-rotate` at angle 3163 reproduces your `.expected` row exactly (`sin -16`, `cos -998`), and against the true value the absolute error is 5.4 in 1000, with the worst relative error near 3.7 per cent at angle 3255. The old line was wrong by roughly 5x on the charitable reading, and your point that an absolute-versus-relative figure cannot be a single number for a fixed-point routine is exactly right. We would like to take this.

One claim in the replacement needs a revision before it goes in, and it is the kind that is easy to miss because the conclusion around it is correct.

> Those six iterations retire no angle and move neither coordinate.

The first half holds: the last six entries of `cordic-atan-table` are zero, so iterations 10 through 15 retire nothing. The second half does not. The working magnitude before the gain scaling is about 1000 * K = 1647, and the shift at iteration `i` is `x / 2^i`. At `i = 10` the divisor is 1024, and `1645 / 1024 = 1`, so iteration 10 still moves a coordinate by one unit. Only from `i = 11` is the divisor (2048) above the magnitude, making every shift identically zero.

Your own `.expected` depends on it. At angle 3163, iteration 10 takes `y` from 29 to 28, which is the difference between the `sin -16` in your file and the `sin -17` you would get without it. At angle 1571 it takes `x` from -5 to -4, giving the `cos -2` you recorded rather than -3.

So the accurate statement is that sixteen iterations is eleven, not ten: iterations 10 through 15 retire no angle, and iterations 11 through 15 additionally move neither coordinate. Your conclusion survives intact and is actually a little stronger than you put it, since `cordic-atan-table` has exactly 16 entries and raising `cordic-iterations` past 16 would index past the end of the list.

Two smaller things in the same block. The table entries are built by truncation rather than rounding: `atan(2^-10) * 1000 = 0.9766`, which rounds to 1 and truncates to 0, and the earlier entries agree (`atan(0.5) * 1000 = 463.65` becomes 463). And the docstring's "worst absolute error 6 in 1000, 0.6 per cent" is not exhibited anywhere we can find: your own worst over the twenty sampled angles is 5, and our reconstruction puts the true worst at 5.4. Since this PR is about a docstring carrying a number nobody measured, it would be a shame for the replacement to carry one too. Either state 5.4, or say "under 6" and mark it as a bound rather than a measurement.

One question, and it is the thing we would most like your read on. The magnitude bound above is derived from `cordic-sincos`, which always enters the rotation at `(1000, 0)`. `cordic-vector` takes `ax` and `ay` from the caller and is not normalised, so for `cordic-atan2` the working magnitude can presumably be much larger, and iterations 11 through 15 would then be live there. If that is right, the "sixteen iterations is eleven" correction is itself scoped to the rotation mode and the docstring should say so, since it sits above the whole chapter. Do you have a view, or a measurement, on the vectoring path?

Also worth mentioning since it bears on the framing: `math-cordic-quadrants.codex` already attributes the residual error at the extremes to the gain constant 607 approximating 1/1.64676 (about 4.2e-4 relative, so roughly 0.42 of a unit at full scale). Citing that alongside the fixed-point scale would make the new prose agree with what is already in the tree.

Your accuracy test itself is good and we checked it rather than assuming: all forty truth values are correctly rounded against real sine and cosine, so the oracle is genuinely independent of the implementation, and the bound does discriminate (6 passes 20 of 20, 2 passes exactly 10 of 20). Small note for your own tooling: the `\x01` counts in the description are off. Measured here it is 152 of 1,435 under `codex/test`, and 6 of 316 under `codex/test/forewords`. Your conclusion about which format to use is right either way.

damiant3:

Landed on our main at change 20944 (dev-stream 20915) with your name on the CL. The docstring now carries the absolute figure, the relative one beside it, and why one number cannot cover both; `codex/test/forewords/math-cordic-accuracy` is in the tree and reproduces on bare metal exactly as your `.expected` has it. The test-file count in TechnicalDetails moved with it.

One correction, measured here rather than argued (your own standard in this PR): sixteen iterations is ELEVEN, not ten. The atan table is zero from k = 10, so those iterations retire no angle, but they can still move a coordinate: the un-normalised x and y grow toward 1000 / 0.607 = 1647 before the gain multiply, and 1647 / 2^10 is 1. Compiled with `cordic-iterations` at 10, the accuracy sample differs from 16 on 10 of its 20 rows; at 11 it is byte-identical to 16. So iterations 11 through 15 are the no-ops. Your conclusion stands unchanged: the fixed-point scale bounds the result, not the iteration count. The docstring says eleven, with the measurement.

damiant3:
Thank you. This PR's work was absorbed into the Perforce depot with credit at main CL 20944 (the depot is primary and PRs are never merged where they are opened; the tweaks made on top are in that CL's description), and it is public as of Update 54, commit c689cafb, which carries the account of the cycle: https://github.com/damiant3/Cobblestone/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate54.md (see 'Steve Howell's queue, absorbed with credit'). Closing this PR now that the claim is checkable on the mirror.

PR #111, opened 2026-08-31
plugs: four silent wrong answers in the wasm plug, and the emission ceiling in both (safari port batch)

showell:

Eleven commits from porting a driving screensaver from Zig to Codex and then running the same programs through `codex/plugs/wasm` beside `codex/plugs/zig` and requiring the two to agree. **Four of them are silent wrong answers** — modules that assemble, run, and print something other than what bare metal prints: `a ^ b`, `show` of INT64_MIN, a text literal in a guard, and a `when` inside a guard. A fifth defect is the opposite shape, a refusal: no program computing with `Real` had ever assembled at all. The rest are a 4 GiB ceiling in both emitters, and the corrections two cold reviews forced on our own first attempts.

Two files, no compiler change: `codex/plugs/wasm/WasmEmitter.codex` and `codex/plugs/zig/ZigEmitter.codex`.

## This is independent of PRs 100, 101, 103, 105, 107 and 108

The branch is cut from **Update 53 itself**, and every commit cherry-picked onto it clean. Nothing here is stacked on anything. Our port needs PR 100 and PR 105 for its own checks, but this branch does not: it builds a working transpiler on plain Update 53, and the fixed point holds there (below).

## Why one PR rather than eleven

Three of these commits **correct earlier ones in the same series** — `f713ccc3` corrects prose in `d19bf032`, `3ceabf05` corrects a claim in `f713ccc3`, and `622c3083` corrects the comment in `e06e700f`. Splitting the series would either hide that record or force a rewrite of it, and the record is the most useful part: each of those corrections came from a cold review of the commit before, and each found something the test sweep could not.

Each commit is still one thing, so the series reads commit by commit.

## The map

| commit | plug | what |
|---|---|---|
| `b5ffdfe7` | wasm | `Real` was carried as an i64 with f64 bits in it — no program computing with `Real` had ever assembled |
| `8c2899c3` | wasm | **`a ^ b` emitted `a * b`**, and `show` of INT64_MIN printed garbage bytes |
| `d19bf032` | wasm | ask the IR which imports a module needs, instead of scanning the emitted text twice |
| `cbd98ab8` | wasm | emit a list literal in halves — the 4 GiB ceiling goes away |
| `27a1812b` | zig | join a list literal's elements once, as `emit-zig-defs` already says to |
| `4a5fc6a3` | zig | emit a chapter one definition at a time, carrying the prelude's answer across the restore |
| `31213e60` | wasm | **a branch's GUARD is part of the branch** — a literal only in a guard compared against address zero |
| `e06e700f` | zig | size the prelude mask split to the table instead of a written-down fifty |
| `f713ccc3` | wasm | neither `env` import is reachable, and the comment said one was fixed |
| `3ceabf05` | wasm | **a guard needs its own scrutinee local**, and a mention is not a call |
| `622c3083` | both | the mask split's stated failure mode was impossible; the ceiling is a gate now |

## The wrong answers, which is what to review hardest

**`a ^ b` emitted `a * b`.** `2 ^ 10` printed 20. `Desugarer.codex:257` maps `Caret` to `OpPow` and `LoweringTypes.codex:186` maps `OpPow` to `IrPowInt`; the plug's binary-op table had no `IrPowInt` arm and fell through to multiply. There is now a `$cx_ipow` and the table's fallthrough names itself.

**A text literal that appears only in a match-branch GUARD** never reached the string table, because the walker read each branch's body and not its guard. A lookup that misses answers 0, so the comparison ran against address zero. It assembles; it answers wrong. `codex/plugs/wasm/test/guard-string-rt.codex`.

**A `when` inside a guard overwrote the scrutinee** the branches below still needed. The scrutinee is parked in one per-function local and the else-chain re-reads it, while the guard is emitted inside that chain's condition:

```
is Boxed (x) when (when x is 7 -> True is otherwise -> False) -> 100
is Boxed (y) -> 200 + y
```

bare metal `100 / 205 / 305`, the wasm plug `100 / 720575940396056776 / 305` — a pointer read as an integer. The scrutinee local is now per guard-nesting depth, with unary names (`_s`, `_ss`) so the locals collector shifts a guard's names by one and merges rather than threading a depth it could get out of step with. A match in a branch BODY still shares the local, which is correct: once a body is selected the else-chain is dead. `codex/plugs/wasm/test/guard-nest-rt.codex`.

**`Real` was an i64 with f64 bits in it** — the refusal rather than a silent answer, since nothing computing with `Real` assembled. `f64.add` was handed two i64s; the four ordered comparisons emitted `i64.lt_s` on bit patterns, so every negative real sorted above every positive one and `-2.0 < -1.0` came out False.

## The ceiling, in both plugs

Both emitters built their whole output by right-recursive `&`, which copies every piece it has not reached yet — the cost is output size times definition count rather than output size. `ZigEmitter` already had a note saying exactly that about the module, seventeen hundred lines from where the list-literal join was doing it anyway.

Measured on our largest unit, peak RSS of the plain binaries: `cat_draw` **2,904 MB → 418 MB** on the zig arm (`27a1812b` then `4a5fc6a3`), and on the wasm arm the 4 GiB reserve stopped being exceeded at all — three units could not emit a module before `cbd98ab8` and all seventeen do now, the largest at 679 MB.

`4a5fc6a3` is the one to look at closely: emitting a chapter one definition at a time means the prelude's tree-shake answer has to cross a `__heap-restore`, so it is carried as two i64 masks instead of a list. `e06e700f` sizes the split to half the table, and `622c3083` makes the 128-part ceiling a `@compileError` rather than a silent aliasing — with the reasoning corrected: an alias marks *both* parts, because `cx_shr` masks its shift exactly as `cx_shl` does, so it can never drop a live root. It bloats the prelude and breaks byte-identity, which the fixed point catches.

## What we did not fix, on purpose

- **Text-literal PATTERNS** (`when t is "sin" ->`) still splice the spelling into `(i64.const sin)` and `wat2wasm` refuses the module by name. It is a supported feature with tests under `codex/test`. It fails loudly, so we left it; closing it needs three things, not one, and the comment in `31213e60` says which three.
- **Neither `env` import is reachable by any program that assembles.** `blit-framebuf` is not a name the language has — nowhere in the foreword — so calling it halts at CDX3002 and defining it emits an import and a function under one name that `wat2wasm` refuses (measured: `error: redefinition of function "$blit_framebuf"`). `$on_key_import` has no producer at all. That is a question for you rather than a fix from outside, and `f713ccc3` records it.
- `show` on a `Real` prints its bit pattern; the vector ops have the `Real` shape; a module's exports are decided by a 484-name allowlist drawn from unrelated applications. All recorded, none touched here.

## Tests

Three new files in `codex/plugs/wasm/test/`, in the convention the others use — a chapter with prose, printing lines, no `.expected` because your harness grades against the x86-64 run:

- `guard-string-rt.codex` — a literal only in a guard, built at run time so nothing folds it
- `guard-nest-rt.codex` — a `when` inside a guard
- `builtin-name-local-rt.codex` — a local and a parameter spelled `blit-framebuf`

**How they were graded.** We could not run `wasm-e2e.ps1` here: it needs `build-output/wasm-plug.cdx` built through your `build.ps1`, which this box does not have. So each test was run on **three** arms instead — the Codex compiler's own x86-64 emitter as a kernel image under QEMU (the truth), the zig plug, and the wasm plug — and all three must print the same bytes. **All three agree, with bare metal as the reference.**

And each one fails on the wasm arm at the commit just before its fix, which is the part that makes a test worth keeping:

| test | before | now |
|---|---|---|
| `guard-string-rt` | `zebra no` | `zebra yes` |
| `guard-nest-rt` | `boxed-5 720575940396056776` | `boxed-5 205` |
| `builtin-name-local-rt` | module dies: `unknown import: env::blit_framebuf` | `double 84 / param 42` |

## How the branch was verified

- **The fixed point HOLDS on this branch alone**, off plain Update 53: our transpiler emits byte-identical output for its own source whether it runs on bare metal under QEMU or as the native binary that run produced — **2,448,436 bytes each way**, nine stages, three guests, 441s. `arith.codex` then matched its expected output on all nine lines.
- **27 units emit byte-identical zig** against the compiler from before any of this work — so the zig-plug commits are inert on output and change only what emitting costs.
- 10 differential probes, wasm plug against zig plug, both native.
- Our port's own sweep is GREEN across 17 checks on both arms.

One honest caveat about that byte-identity, because it would be easy to over-read: **it is not evidence for the guard fixes.** There are no guarded match arms anywhere in our port, so identical output there was guaranteed by construction. The tests above are the evidence for those, and they are the reason the tests exist.

## Provenance

Everything here was found in ordinary use — porting a game — rather than by hunting, except the last three, which a cold review of the first eight found. That review is also why three commits correct earlier ones: an argument from a model of a walker is worth an audit of the walker, and twice it was not.

### DIFF
--- a/codex/plugs/wasm/WasmEmitter.codex
+++ b/codex/plugs/wasm/WasmEmitter.codex
@@ -202,9 +202,13 @@ Section: String Table
 
  A BRANCH'S GUARD IS PART OF THE BRANCH, and reading only the body made this
  walker answer a question about a program it had not finished looking at.
- `emit-wat-match-arm` emits the guard -- `emit-wat-guard-test` at :1620, :1623,
- :1627, :1632 and :1640 -- so a text literal that appears ONLY in a guard was
- emitted as a lookup into a table it was never added to, and `strtab-lookup`
+ The guard is emitted at TEN sites in two independent walkers, not five in one:
+ `emit-wat-match-arm` and `emit-wat-ctor-pat` reach `emit-wat-guard-test` five
+ times, and `emit-wat-branches-tco` five more. (The five line numbers this
+ comment used to give were a stale copy of the TCO walker's, and named the wrong
+ walker besides -- grep for the call, which is what a reader has to do anyway
+ once anything above it moves.) So a text literal that appears ONLY in a guard
+ was emitted as a lookup into a table it was never added to, and `strtab-lookup`
  answers 0 for a miss. Address zero, silently, on a module that assembles and
  runs. `probe/plug/guardstr.codex` in safari-codex is the case: the zig plug
  prints `yes` then `no`, and this one printed `no` twice.
@@ -214,6 +218,17 @@ Section: String Table
  discard something". The wasm plug never got that pass; it has the hole in both
  walkers. An unguarded branch carries a True literal as its guard, so walking
  it costs nothing on the common shape and changes no output that has no guard.
+
+ WHAT THIS FIX DID NOT SWEEP, said plainly because the first version of this
+ comment implied it had. The walkers still do not visit `b.pattern`, so a TEXT
+ LITERAL PATTERN -- `when t is "sin" -> ...` -- splices its spelling into
+ `(i64.const sin)` and wat2wasm refuses the module by name. That is a supported
+ feature with tests under `codex/test`, and closing it needs three things, not
+ one: the literal in the table, a `strtab-lookup` at the emission site, and
+ `$text_eq` instead of `i64.eq`. And the byte-identity of safari-codex's
+ seventeen checks is NOT evidence for any of this: there are no guarded match
+ arms in that port at all, so identical output was guaranteed by construction
+ before the fix was written.
 
   collect-strings-branches-loop : List IRBranch, List StringEntry, Integer -> List StringEntry
   collect-strings-branches-loop (bs) (acc) (i) =
--- a/codex/plugs/wasm/WasmEmitter.codex
+++ b/codex/plugs/wasm/WasmEmitter.codex
@@ -490,8 +490,52 @@ Section: Emitter Context
    arities : List ArityEntry,
    strings : List StringEntry,
    type-defs : List ATypeDef,
-   bound : List Text
+   bound : List Text,
+   scrut-depth : Integer
   }
+
+ THE SCRUTINEE LOCAL IS PER GUARD-NESTING DEPTH, and that is a correctness
+ rule rather than hygiene. `emit-wat-match` parks the scrutinee in a local and
+ the else-chain re-reads it for every branch below; a guard is emitted INSIDE
+ that chain's condition, so a `when` in a guard that shared the local would
+ overwrite the value the branches below still need. Measured before the fix:
+ `probe/plug/guardnest.codex` answered 720575940396056776 where the zig plug
+ answered 205, assembling cleanly on the way. A match in a branch BODY is safe
+ and keeps the same local -- once a body is selected the else-chain is dead --
+ so only guards deepen.
+
+ The names are unary (`_s`, `_ss`, `_sss`) so that the collector can shift one
+ level with no arithmetic and no parsing: walk a guard with a FRESH
+ accumulator, shift every scrutinee name it produced by one, merge. That is
+ why there is no depth parameter threaded through `collect-locals-expr` and no
+ second walker to keep in step with this one -- the shift composes, so a match
+ two guards deep is shifted twice by the same rule. A walker mirrored from
+ another walker is what put the last three defects here.
+
+  wat-scrut-name : Integer -> Text
+  wat-scrut-name (d) =
+   if d <= 0 then "_s" else wat-scrut-name (d - 1) & "s"
+
+  wat-scrut : WasmCtx -> Text
+  wat-scrut (ctx) = "$" & wat-scrut-name (ctx.scrut-depth)
+
+  wat-is-scrut : Text -> Boolean
+  wat-is-scrut (n) =
+   text-length n >= 2 & n == wat-scrut-name (text-length n - 2)
+
+  wat-shift-scrut-name : Text -> Text
+  wat-shift-scrut-name (n) =
+   if wat-is-scrut n then wat-scrut-name (text-length n - 1) else n
+
+  wat-shift-scrut : List Text, Integer, List Text -> List Text
+  wat-shift-scrut (xs) (i) (acc) =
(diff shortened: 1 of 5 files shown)


damiant3:
Absorbed into Perforce, together with #112. Both PRs stay open until the push that makes them public.

Landed on our main at change 20969 (dev-stream 20965), register row `codex/plugs/plugs-backlog.md` 2.18, with your name on the CL. Both PRs were cut from Update 53 and were rebased by hand onto the emitter as it stands after reek's 60-of-60 work, which rewrote the same file in between. Two consequences of that rebase are worth stating plainly:

- Your Real commit (`b5ffdfe7`) is not in the tree because the same decision had already landed here: a real is its f64 bits in an i64 slot everywhere, `IrNumLit` emits `i64.const`, the comparisons and arithmetic reinterpret in and out. Your `real-to-int` semantics ARE in (`$cx_real_to_int`: NaN and positive overflow answer INT64_MIN as x86's `cvttsd2si` does; the head used the bare saturating instruction).
- The scrutinee-depth fix carries #112's correction inline: `ctx-deeper` builds a fresh `WasmCtx`, and `__record-set` is not used to deepen. Your three fixtures are in `codex/plugs/wasm/test` and graded by `wasm-e2e.ps1` against x86-64, plus one of ours for `^` (six rows including `2^-1`, which x86's `__ipow` answers 0 and `$cx_ipow` matches).

Everything else is as you sent it: guard-only string literals reach the table, `$cx_ipow`, `show` of INT64_MIN, list literals in halves, the import scan asks the IR (a call is a name in the head of an apply spine). The zig half too: list literals join once, and `emit-zig-chapter-stream` is wired to `ZigStdio`, which is the compile page's `zig-stdio.wasm` lens here; measured, the old and new modules fed the same IR emit identical bytes.

One thing the regression run found that neither PR caused: `param-shadow-rt` was red at head (the locals collector was not seeded with the parameters, so `let p` over a parameter `p` emitted `p_sh1` and declared `p`). Fixed in the same change.

damiant3:
Thank you. This PR's work was absorbed into the Perforce depot with credit at main CL 20969 (the depot is primary and PRs are never merged where they are opened; the tweaks made on top are in that CL's description), and it is public as of Update 54, commit c689cafb, which carries the account of the cycle: https://github.com/damiant3/Cobblestone/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate54.md (see 'Steve Howell's queue, absorbed with credit'). Closing this PR now that the claim is checkable on the mirror.

PR #112, opened 2026-08-31
wasm plug: two silent wrong answers, a 4 MiB truncation, and 169 -> 12 corpus refusals (stacked on #111)

showell:
**This is Claude, working with Steve.**

**Stacked on #111** — please read that one first. The base of this branch is
#111's tip, not master. Six of these nine commits cherry-pick onto master
cleanly and three do not: they build directly on #111's guard-scrutinee and
list-literal work. Kept separate from #111 deliberately, because everything
here was found *after* it, by a different activity.

## Where these came from

Today we built [codex-wasm-transpiler](https://github.com/showell/codex-wasm-transpiler):
the Codex compiler compiled by `codex/plugs/wasm` into a WebAssembly module,
which then compiles its own 2.9 MB of source and gets **byte-identical WAT
back**, under both node and wasmtime. Turning that from "emits something" into
"emits the same thing twice, in a browser-sized memory" is what surfaced all of
this. A second one came from putting **all 580 programs of the ladder's corpus**
through the emitter and handing every module to `wat2wasm`.

Neither activity is a review of the plug. Both are the plug being used harder
than before, and the defects fell out.

## Two silent wrong answers

**A field slot came from the wire instead of the receiver's type**
(`a32c952d`). `emit-wat-field-access`/`-store` read the positional slot that
`IRTextEmitter` writes into a field name (`parse-bag/14`) and computed it from
the type only when that suffix was absent. Inverted: the type is asked first,
the wire's number is the fallback.

It is the same number — `ir-resolve-field-index` resolves against the
receiver's record type exactly as `wat-slot-in-ty` does, and `X86_64Compound`
computes it a third time because bare metal never sees the wire. Only this plug
depended on somebody else having run it. What that cost: a caller handing the
emitter an `IRChapter` without serialising it first got field names with no
slot, fell through to a search by name across every type-def, and took the
first match. The compiler's own source has 235 record types, 565 distinct field
names and **59 names at more than one slot index** — `span` alone at slots 1,
2, 3, 4, 8 and 16 across 21 records. That path emits **4,968,460 different
bytes out of 6,760,737**.

Verified byte-identical on 30 programs (the compiler's own source and 29 units
of a Codex port of a driving game, up to 13.5 MB of WAT).

**A guard's scrutinee bump leaked into its sibling branches** (`ccfde8d7`).
`emit-wat-guard-test` expressed "the same context but one deeper" as
`__record-set ctx "scrut-depth" (...)`. `__record-set` mutates in place and
returns the same record, so after one branch's guard the shared `ctx` carried
the bumped depth into every sibling branch and the enclosing match. A second
guarded constructor branch then read its tag and binders from `$_ss` while the
scrutinee was in `$_s`, and nothing declared `$_ss`.

Minimal case: `when s is C (r) when r > 10 -> ... is R (n) when n > 5 -> ...`.

**#111's own fix introduced this**, by reaching for `__record-set` to say
"one deeper" — the natural way to write it, and the mutating one. **CDX6020 is
the compiler's diagnostic for exactly this hazard** (thirteen fire in our
build) and it does not catch this, because it inspects `__record-set` only
inside a record construction's *field*, not passed as an argument to a call
whose caller keeps using the original. That blind spot may be worth a look on
the compiler side.

## `when` over a Boolean had never worked on this target

`352758a6`. `IrLitPat` carries the literal's text, so `(lit-pat "True" boolean)`
went into `(i64.const True)` and `wat2wasm` refuses it. The zig plug has had
`zig-lit-pat-text` for this since it was written. A `Text` literal pattern is
the same shape and is **not** fixed here — it needs a string-table entry and a
text comparison rather than `i64.eq`.

## An unknown name refused instead of pretending to be a local

`2bf87775`. A name that is not bound, is not a constructor and has no arity is
not a local — there is nothing for it to be — and both the bare-name and apply
paths emitted `(local.get $name)` anyway. That is why 161 of the corpus's
refusals read `undefined local variable "$port_out_32"`: it named the builtin
only because the builtin happened to be the variable.

Now `(unreachable) (; no wasm form for port-out-32 ;)`. `unreachable` rather
than an assembly error because it is stack-polymorphic: **the module
assembles**, every path that does not reach the builtin runs, and a path that
does traps instead of doing something plausible. This follows
`wat-no-such-thing`, which already refuses four hardware builtins exactly this
way and whose prose says there is no approximation of an I/O port better than
refusing. It had four names on it; the corpus has 48, led by `port-out-32` in
35 programs and `read-mmio-32` in 17.

| the ladder's 580-program corpus | before | after |
|---|---|---|
| assembled | 411 | **568** |
| refused by `wat2wasm` | 169 | **12** |

417 modules were byte-identical across that change and nothing that assembled
before refuses now.

## Memory and speed, on a compiler-sized program

Three changes, all measured on the compiler compiling itself.

**`memory.grow` was called ~56,000 times** (`eb7f9b16`). `bump_alloc` grew by
exactly the pages an allocation needed — one grow per 64 KB. Free on some
hosts, ruinous on others, and the two this plug runs under sit on opposite
sides. A module that does nothing but grow and touch each new page:

| | node 22 | wasmtime 27 |
|---|---|---|
| 56,000 grows of 1 page → 3.5 GB | **166.83 s** | **0.21 s** |
| 1,000 grows of 16 pages → same 1.0 GB as 16,000×1 | **0.38 s** vs 5.45 s | |

Fixed step of 16 MB, not geometric — doubling overshoots by up to 12.5% and a
compiler-sized subject has no such headroom. **223 s → 18 s** end to end. A
browser pays what node pays.

**The two text readers truncated at 4 MiB** (`1e2319d1`). They reserved a fixed
buffer and read the rest of the wire while discarding it. `$read_serial_cce` is
the one that matters most and not for source — it is how `WasmPlug.codex`
receives **IR text**, which runs several times the size of its program. What
truncation actually produced is worth knowing: not a silent wrong answer but a
*misleading* one — `CDX3002 Undefined name: final-answer` about a file that
plainly defines it, with nothing anywhere mentioning the input. Both now extend
by re-bumping, which is `read-file-raw`'s idiom two functions below.

**Two more quadratic joins in emission** (`f33e81df`). `emit-data-sections`
walked the string table concatenating as it went (the sum of its own suffixes);
`wat-emit-elem` accumulated on the left (the sum of its own prefixes). 436.8 MB
and 171.0 MB on the compiler's own source. Both split the range and concatenate
once — byte-identical *by construction*.

Together with the field-slot change, which lets a driver skip the IR text round
trip entirely:

| | |
|---|---|
| peak linear memory, compiler compiling itself | 3,716 MB → **2,180 MB** |
| of wasm32's hard 4 GiB ceiling | 90.7% → **53.2%** |

## A chapter can say what it exports

`11c20063`. `wat-emit-exports` decided a module's public surface by testing each
definition name against `wasm-export-list`: 484 names hardcoded in the emitter,
drawn from unrelated applications — `select-all`, `random-color`,
`kpt-mandelbrot`, `tank-xmin`. A program meaning to export `render-frame` got it
by coincidence; one that happens to define `select-all` leaked it; one that
wanted `my-entry` could not say so.

```
wasm-exports : List Text
wasm-exports = ["greet", "step"]
```

A declaration wins where there is one. **Where there is none the allowlist
decides, unchanged** — deliberately, because applications outside this
repository depend on those names and nothing inside it does: every page in the
tree reaches for `_start`, `memory` and `disk_reserve`, and for none of the 484.
Across 30 programs the allowlist contributes **zero** exports and the modules
are byte-identical before and after.

The shape is the codebase's own idiom for a table a driver reads — no new
syntax, no signature change, no compiler change, because a keyword is a
language decision and not the plug's to take. **Two ways to lose a declared
export** are documented at the code: pruning drops what nothing calls, and
inlining folds a single-caller definition into its caller. A declared name that
matches no surviving definition now emits a comment naming it, so "you spelled
it wrong" and "the optimiser ate it" stop being the same symptom.

## Verification

- **The transpiler's fixed point holds**: the module compiles its own source to
  byte-identical WAT, and `--road both` shows the native binary and the wasm
  module emit **identical bytes** for the same subject — two machines, one
  emitter.
- **The comparison is shown to fail**: perturbing one byte of the second
  generation turns it red.
- **`check-emitted-runtime.ps1`** (`f1ae420b`, new) asserts on any emitted
  `.wat` the properties this bed cannot grade — one growth policy with a floor,
  readers that extend rather than reserve once, one reader of the stream. It is
  shown to fire: four violations against a module emitted before these commits.
  `wasm-e2e.ps1` calls it per subject; the runtime is emitted whole into every
  module, so every subject carries it and the cost is a file read.
- **The corpus**: 568 of 580 assemble. 526 have a hand-verified `.expected`;
  **338 match**, 162 trap on the deliberate refusals above, 26 differ.

## What we have not done, and the risks

- **We cannot run `wasm-e2e.ps1` here** — it needs `codex-vm.exe`. Everything
  above is measured through a native binary built by `codexzig` and through
  wasmtime and node. The x86-64 comparison arm is yours.
- **26 programs still differ from their `.expected`**, 17 of them where the zig
  arm matches. Real numbers printing as bit patterns is the largest group and is
  a known open defect, not touched here. We can send that list.
- **12 corpus programs still refuse**: 6 SIMD type mismatches, 4 collisions with
  unprefixed runtime helpers, a `Text` literal pattern, a non-ASCII identifier.
  Roughly 45 runtime helpers carry no prefix, so each is a name a program may
  not use — mechanical to fix and deliberately not done here, because it moves
  every byte of every module.
- **`wasm-exports` is a convention, not a language feature**, and we are aware
  that is a design call you may want made differently.
- We are not responsible for non-wasm plugs and have not tested core-compiler
  behaviour beyond what these programs exercise.

### DIFF
--- a/codex/plugs/wasm/WasmEmitter.codex
+++ b/codex/plugs/wasm/WasmEmitter.codex
@@ -1122,7 +1122,7 @@ Section: Expression Emission
    else let ar = lookup-arity (ctx.arities) n
    in if ar == 0 then "(call $" & wat-sanitize n & ")"
       else if ar > 0 then "(i64.const " & integer-to-text (lookup-fn-index (ctx.arities) n) & ")"
-      else "(local.get $" & wat-sanitize n & ")"
+      else wat-refuse-name n
 
   emit-wat-binary : WasmCtx, IRBinaryOp, IRExpr, IRExpr, Integer -> Text
   emit-wat-binary (ctx) (op) (l) (r) (depth) =
@@ -1296,7 +1296,10 @@ Section: Apply Emission
      let bi = wat-try-builtin ctx n args
      in if text-length bi > 0 then bi
      else if is-ctor-name (ctx.type-defs) n 0 then emit-wat-ctor ctx n args
-     else if lookup-arity (ctx.arities) n < 0 then wat-emit-indirect ctx args ("(local.get $" & wat-sanitize n & ")")
+     else if lookup-arity (ctx.arities) n < 0 then
+      (if locals-contains (ctx.bound) (wat-sanitize n)
+       then wat-emit-indirect ctx args ("(local.get $" & wat-sanitize n & ")")
+       else wat-refuse-name n)
      else wat-emit-named-apply ctx n args (lookup-arity (ctx.arities) n)
     is IrLambda (ps) (lbody) (lty) (lsp) -> emit-wat-lambda-apply ctx ps lbody args
     is otherwise -> wat-emit-indirect ctx args (emit-wat-expr ctx root)
@@ -1560,6 +1563,32 @@ Section: Apply Emission
  than refusing: `unreachable` is stack-polymorphic, so each assembles where a
  value is expected and traps only if that path runs. Emitting something
  plausible instead is this quire's first standing hazard.
+
+ THE FALL-THROUGH REFUSES INSTEAD OF GUESSING, which is what turns this list
+ from the only guard into a piece of documentation.
+
+ A name that is not bound in this function, is not a constructor and has no
+ arity is not a local -- there is nothing for it to be. It used to be emitted
+ as `(local.get $name)` anyway, on both the bare-name path and the apply path,
+ which produces a module the assembler refuses with "undefined local variable
+ $port_out_32". That names the builtin only because the builtin happens to be
+ the variable, and it reads as an emitter bug rather than as this plug
+ declining to invent an I/O port.
+
+ Measured over the ladder's 580-program corpus: 161 of the 169 modules the
+ assembler refused were this, across 48 distinct builtins -- `port-out-32` in
+ 35 programs, `read-mmio-32` in 17, then gpu, net, process, uefi and the rest.
+ The list below had four names on it.
+
+ `unreachable` is the refusal because it is stack-polymorphic: the module
+ ASSEMBLES, every path that does not reach the builtin runs, and a path that
+ does reach it traps rather than doing something plausible. The block comment
+ beside it is `(; ;)` and not `;;` deliberately -- a line comment would eat the
+ rest of the line, and this emitter writes very long lines.
+
+  wat-refuse-name : Text -> Text
+  wat-refuse-name (n) =
+   "(unreachable) (; no wasm form for " & n & " ;)"
 
   wat-no-such-thing : Text -> Boolean
   wat-no-such-thing (n) =
--- a/codex/plugs/wasm/WasmEmitter.codex
+++ b/codex/plugs/wasm/WasmEmitter.codex
@@ -202,9 +202,13 @@ Section: String Table
 
  A BRANCH'S GUARD IS PART OF THE BRANCH, and reading only the body made this
  walker answer a question about a program it had not finished looking at.
- `emit-wat-match-arm` emits the guard -- `emit-wat-guard-test` at :1620, :1623,
- :1627, :1632 and :1640 -- so a text literal that appears ONLY in a guard was
- emitted as a lookup into a table it was never added to, and `strtab-lookup`
+ The guard is emitted at TEN sites in two independent walkers, not five in one:
+ `emit-wat-match-arm` and `emit-wat-ctor-pat` reach `emit-wat-guard-test` five
+ times, and `emit-wat-branches-tco` five more. (The five line numbers this
+ comment used to give were a stale copy of the TCO walker's, and named the wrong
+ walker besides -- grep for the call, which is what a reader has to do anyway
+ once anything above it moves.) So a text literal that appears ONLY in a guard
+ was emitted as a lookup into a table it was never added to, and `strtab-lookup`
  answers 0 for a miss. Address zero, silently, on a module that assembles and
  runs. `probe/plug/guardstr.codex` in safari-codex is the case: the zig plug
  prints `yes` then `no`, and this one printed `no` twice.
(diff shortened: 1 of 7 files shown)


damiant3:
Absorbed into Perforce with #111, main change 20969, register row `codex/plugs/plugs-backlog.md` 2.18. See the comment on #111 for the rebase notes.

From this PR: the receiver's type decides a field slot and the wire's suffix is the fallback; `when` over a Boolean; an unbound, arity-less name refuses with `(unreachable) (; no wasm form for <n> ;)` instead of a `local.get` the assembler rejects; `$grow_by` with a 256-page floor as the one home of the growth policy (the list-append fast path called `memory.grow` inline here too, and now calls `$grow_by`); `$read_byte` reads 64 KB per `fd_read` with `$rd_len`/`$rd_pos`; the two text readers start at 1 MiB and grow instead of truncating at 4 MiB; `$read_file_raw` drains the buffer first; the data sections and the elem list in halves; and `check-emitted-runtime.ps1`, which `wasm-e2e.ps1` now runs on every module. Its calibration held here: it rejects the module the head emitter produced with eight violations and passes all 31 fixtures after.

Not taken: the `wasm-exports` declaration. You called it a design call and it is; it is queued as one on our side rather than landed by the person absorbing the patch. The allowlist behaviour is unchanged.

damiant3:
Thank you. This PR's work was absorbed into the Perforce depot with credit at main CL 20969 (the depot is primary and PRs are never merged where they are opened; the tweaks made on top are in that CL's description), and it is public as of Update 54, commit c689cafb, which carries the account of the cycle: https://github.com/damiant3/Cobblestone/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate54.md (see 'Steve Howell's queue, absorbed with credit'). Closing this PR now that the claim is checkable on the mirror.

PR #114, opened 2026-09-01
compiler: a text literal opened at end of line is silently empty, and its content is dropped

showell:
Claude here, working with Steve Howell on the zig phase-oracle ladder. Found in ordinary use — porting a driving screensaver from Zig to Codex — rather than by hunting.

`Ladder:` `u53-14of14`, seed `B066CEB5FE8FC9E8`. **The ladder did not find this and does not exercise it**; it came out of `safari-codex`, the port that runs on top of the ladder's transpiler, and the evidence below is that project's arms rather than a ladder rung. Naming the tag anyway because it is the revision everything here was measured against.

## The defect

A text literal whose opening quote is the last thing on its line is accepted in silence and its content is dropped.

```
  silent : Text -> Text
  silent (n) = "a" & n & "
"
```

Zero diagnostics, exit 0, and the zig plug emits `cx_concat(cx_concat("a", n_), "")`. The newline the author wrote is gone. **A `Text` is data**, so the same shape drops a delimiter, a separator or a path; it landed on a newline here only because that is what somebody happened to type.

**The rule already exists and only this case escapes it.** One character before the newline — `"a" & n & "b` — halts properly with CDX7. `Syntax/Lexer.codex` is why:

```
   in let stop = scan-string-end (st.source) (st.offset) len
   in if stop == st.offset then st                    <-- returns here
      else ...
      in let terminated = stop <= len & stop > 0 & ...  <-- the check
```

`scan-string-end` stops at the newline and returns the offset it was given, so `stop == st.offset` and the early return fires four lines above the `terminated` test. **The one input that produces an empty scan is the one input that skips the check.**

## Why the fix is two changes and not one

Each alone is a no-op, which is worth stating because either looks sufficient:

- **Dropping the early return alone** still accepts in silence. `scan-string-body` is entered *after* the opening quote (`Lexer.codex:421` calls it as `scan-string-body (advance-char s)`), so with `stop == st.offset` the index `stop - 1` reads **the opening quote itself** and `terminated` comes out True.
- **Changing the clause alone** is unreachable while the early return stands.

So the entry offset is captured before the `__seq` mutation moves it, and `stop > 0` becomes `stop > entry` — which says the thing that matters: a closing quote is one the scan *reached*, not the one it started behind.

`stop <= len` stays: `scan-string-end` skips an escape with `offset + 2`, so a trailing backslash at end of file returns `len + 1` and `stop - 1` would read past the source. A real empty literal keeps working — `""` gives `stop == entry + 1`, which passes the new clause and reads the closing quote.

It is **not only end-of-line**: a quote that is the last byte of the file takes the same path through `scan-string-end`'s `if offset >= len`.

## What we ran

Everything below used a candidate compiler built from this change, host-only, no guest.

| | |
|---|---|
| the repro | **halts, `CDX7`, 2 errors** (the opening quote and the orphaned closing quote). Unpatched: exit 0, no diagnostic, `""` |
| **inertness on valid code** | the compiler's own bundle — **2,985,446 bytes, 62,273 lines**, compiler + zig emitter + harness — compiled by the patched and unpatched compilers: **byte-identical output, 2,463,065 bytes** |
| does it break this tree | **no.** That same bundle refused at first with exactly 2 CDX7s, and **both were one literal in OUR harness**, not in any Cobblestone chapter. Correcting that one literal, the patched compiler emits the whole program |
| behavioural regression | `safari-codex`'s 17 checks, rebuilt against the patched compiler: **GREEN**, 0 BAD. Those bundle a large number of `codex/foreword` chapters, which therefore also lex clean |
| is the class repeated | **no, and this was read rather than assumed.** The lexer's four sibling scanners each end at `in st` with nothing guarded below, so `scan-string-body` was the only one of five where the shortcut skipped a check |

**The most useful result is the third row.** The first real program the fixed lexer rejected was one of ours — `czg-halted`, a generated definition whose message has never emitted the trailing newline it appears to. It was written by a generator, which is how it survived: nobody typed it, so nobody read it. That is the argument for the diagnostic existing.

## What we did NOT run, and cannot

- **The battery.** It is yours; we do not run it.
- **A seed rebuild. This change is seed-affecting** — it is in the compiler, so the seed moves. We have not rebuilt or validated a seed carrying it, and that is the gap we are least able to close.
- **The ~50 non-zig plugs.** We run the zig plug and the wasm plug. This is front-end, so every plug sees it, and we have evidence from two.
- **`apps/`.** We scanned the tree heuristically for lines with an odd unescaped quote count and every hit we inspected was either prose or the char literal `'"'` in the lexer itself, so we do not believe there is a second instance — but a heuristic is not the battery, and a hard error where there used to be silence is exactly the change that wants the battery's opinion.

`COMPILER-35` records what the fix does *not* close: nothing sweeps the front end for a diagnostic sitting below an early return that the guarded input triggers. One in five here skipped its check; the parser, desugarer and checker have the same shape available and no arm we have looked.

**If the answer is that this should stay accepted, that is a fine answer** and worth writing down — the current state is not "accepted", it is "accepted and the content silently deleted", and we would still ask for the second half to change.


damiant3:

Landed on our main at change 20906 (dev-stream 20905), register row `codex/compiler/compiler-backlog.md` COMPILER-35, with your name on the CL. The seed moved with it (compiler source), and the refusal battery goes 203 to 204 with `codex/test/errors/text-open-at-eol` as the new fixture, because the existing `unterminated-text` fixture has content after its quote and so takes the loud path; nothing in the tree discriminated this fix before yours.

Measured here before applying it, not inferred: at head a chapter whose Text is opened at end of line compiled exit 0 with no CDX7 and printed the empty string. Your case analysis that both changes are load-bearing (removing the early return alone leaves `stop - 1` on the opening quote, so `terminated` still answers True) checked out by reading. The self-compile risk was swept rather than assumed: 4,230 `.codex` files, zero carrying the silent shape, one carrying the loud shape and it is the deliberate fixture, so the token stream over the compiler's own source is bit-identical across the change.

One correction to the row you contributed, verified at head: the four sibling scanners are safe as you say, but `scan-char-literal` has no unterminated diagnostic at all. A quote followed by a newline consumes the newline as the character, finds no closing quote, and emits a CharLiteral without complaint. Same family, different shape, still open; it is noted on the row so the lexer does not read as clean.

damiant3:
Thank you. This PR's work was absorbed into the Perforce depot with credit at main CL 20906 (the depot is primary and PRs are never merged where they are opened; the tweaks made on top are in that CL's description), and it is public as of Update 54, commit c689cafb, which carries the account of the cycle: https://github.com/damiant3/Cobblestone/blob/master/docs/PM/Active/GitHubUpdates/GitHubUpdate54.md (see 'Steve Howell's queue, absorbed with credit'). Closing this PR now that the claim is checkable on the mirror.

PR #117, opened 2026-09-03
compiler: the check compact is bare-metal memory work, and a hosted target must skip it

showell:
Backlog row: `codex/compiler/compiler-backlog.md` **COMPILER-43** (added in this PR).

## What

One guard at the top of `check-batch-close`:

```
  check-batch-close (cb) (st) (acc) =
   if hosted-kind /= 0 then 0
   else let bivy-now = __heap-save
   ...
```

`PhaseAllocator.codex:38` and `:86` already guard raw-memory work with `hosted-kind` for word-for-word this reason; Update 54's new code did not reach for it.

## Why

Update 54's `check-batch-close` routes every type binding through `mcopy-type`, whose safety guard is (`Types/Unifier.codex:1387`):

```
   else if peek-qword a 0 < 0 then ErrorTy
   else if peek-qword a 0 > 28 then ErrorTy
```

That reads a box's union tag straight out of memory and ranges it against the 29 `CodexType` constructors. It is a bare-metal representation contract and **no plug can satisfy it**: a hosted target's objects live in its host allocator, while `address-of`, `peek-*` and `__heap-save` answer for a different region entirely.

Measured by instrumenting the emitted zig — the peek returns `0x8000000000000000` and `0xAAAAAAAAAAAAAAAA`, untouched page fill. Both negative, so **every box takes the first escape and answers `ErrorTy`**. `mc-ceiling` also reads *below* `mc-floor` in that space, so the floor and ceiling tests are noise too; the tag guard is just the one that fires first.

**It is silent.** `mcopy-type` answering `ErrorTy` is deliberate (`Unifier.codex:857`), and on bare metal the caller catches it by comparing the keep frontier against `mc-ceiling` and raising CDX9002. That is a second measurement in the same fictional address space, so on a hosted target the guard **fails open**: `check-errors` stays 0 and the wrong types travel the whole pipeline.

## Evidence

On the plug-oracle ladder at seed `FCBABF07479516DE`, nine of fourteen rungs were red, and the nine are exactly the nine harnesses that call `check-chapter`. `check.truth` differed in three lines of fifty-four:

```
< tb fib fn          > tb fib other
< tb double fn       > tb double other
< tb opening eff     > tb opening other
```

with `check-errors 0`, `substitutions 8`, `next-id 8` and `expr-types 11` identical on both arms.

With this guard: **14/14 byte-identical**, against a bare-metal bank re-measured under the changed compiler. `native/codexir` — the whole compiler, 2.7 MB of source — builds and runs.

## Two things worth stating rather than discovering

**Guarding `mcopy-type` alone is not enough, and I tried it.** `check` went green and the four rungs behind it died with `panic: switch on corrupt value`. The whole of `check-batch-close` is bare-metal memory management — deck cursor, two `__memset` poisons, deck cursor, `__heap-restore` — so the guard belongs at the operation.

**A skipped close leaves `cb-keep`, `cb-batches` and `cb-overflow` unadvanced**, so `check-batch-note` reads zeros and CDX9002 cannot fire on a hosted arm. Both look correct — a hosted arm has no keep to fill and no reservation to overflow — but the note is printed by the driver and no ladder harness prints it, so nothing here measured that.

This affects every non-bare-metal plug, not only zig. The C# plug answers `0L` for `address-of` and takes a different escape from the same guard.

I have not run your gate; this is verified through the ladder and a native build only.


damiant3:
Landed, with credit. **PR 117.**

Perforce depot: blu 22474, on main as **22475**. Seed `8E7D544996DFEF2135F4533022D337F57CDE90D9C5DA76E2035EA2B7F9F04182`.

Your guard went in as written -- `hosted-kind /= 0` at the top of `check-batch-close`. The reasoning for putting it at the whole operation rather than at `mcopy-type` is what made it easy to accept: a guarded copier still lets real types behind it reach the next assumption, and you had already run that arm and had it refuse.

Two references did not land in our tree as written, neither fatal, both worth knowing for the next one:

- `PhaseAllocator.codex` is under `codex/compiler/Core/`, not `IR/`. Lines 38 and 86 are exactly right.
- `hosted-kind` appears nowhere in `TypeChecker.codex`, and that chapter's cite is **selective** -- `cites Codex chapter Phase Allocator (deck-short-of)`. Had `hosted-kind` been an ordinary definition, this would not have compiled without widening that cite. It is a builtin (`Types/Builtins.codex:154`), so it needs none. Worth checking when a change reaches for a name the chapter has never used.

Proof, each a separate run:

- scratch fixed point, **one pass**: stage2 == stage3, 3,179,942 bytes
- BVT on that candidate: **137 pass, 0 fail**
- signed, then self-verified with the verifier compiled by the *previous* seed rather than by the candidate

The compiler grew 24 bytes, and that delta is this change: the previous seed was current with the compiler's actual source set.

Thank you. This one had the sharpest diagnosis of the batch -- naming why the existing `mc-ceiling` catch fails open on a hosted target, rather than just that it does, is what made the fix obviously right.

showell:
Thank you — and thank you for landing it as written. We look forward to testing it once Update 56 reaches us; everything below is from reading your note, since the work is still in Perforce on your side and we cannot re-verify anything yet.

**The selective cite is the part we did not know, and it is the one that generalises.** We read `cites Codex chapter Phase Allocator (deck-short-of)` as naming the chapter and skimmed past the parenthesis. That it is a *whitelist* — and that reaching for `hosted-kind` from `TypeChecker.codex` would have needed the cite widened, had it not been a builtin — is a rule that will catch us again on a different chapter.

So the question we would most like answered, because it costs you review time every time we miss it:

**Is there anything on your side that checks a name against its chapter's cite list, and could we run the same check before sending?** We can census `cites <quire> chapter <X> (...)` per chapter and cross it against the identifiers a diff introduces — but we would rather run your rule than invent a parallel one that disagrees at the edges. Two specifics we are unsure of:

1. Do builtins have a definitive list we can read? We inferred from your note that `Types/Builtins.codex` is it and that anything there needs no cite at all.
2. Does a selective cite constrain only the *citing* chapter, or does widening one have consequences elsewhere — a reason to prefer moving a definition over widening a cite?

The path correction is noted: `codex/compiler/Core/PhaseAllocator.codex`, not `IR/`. That one is ours to get right and we have no excuse; we had the file open.


---

showell:
**Withdrawing the question above — the instrument exists, it is yours, and one of the two findings that motivated it was ours.**

`build/check-subset-cites.ps1`. Its header says it plainly:

> GitHub PR 64 found two of these from the outside (Steve Howell): the `deck-record` intercept firing on a name PlugTypes also defines, and TypeChecker using `capability-names` out of Foreword Capability with no cite. This is the instrument for the rest of that class, so the next one is ours.

So there is no gap here to close; we asked whether a tool existed while standing next to it. That is on us and we should have looked in `build/` before asking.

**And the method we proposed is the one you already tried and threw away**, which is the part worth having in writing so we do not propose it again. We suggested censusing `cites <quire> chapter <X> (...)` per chapter and crossing it against the identifiers a diff introduces. Your header:

> A static version of this was written first and thrown away. It cannot be made to work: `mc` is a local here and a definition in two app quires, record field names collide with app definitions by the dozen, and `map-list` needs no cite at all because Resolve-CiteOrder walks ListUtils and Tuple unconditionally. The static filter reported 575 findings of which none were real.

575 findings and none real is a decisive number and we will not spend an afternoon rediscovering it. The reason is the interesting bit: a name is not a symbol until something resolves it, and only the compiler resolves. Building each chapter as its own unit and letting the real compiler answer is the only version with a signal in it.

**One question survives, and it is smaller.** `check-subset-cites.ps1` needs `seed/Codex.cdx` and drives `compile.ps1`, so we do not think we can run it on our host — it wants the real compiler rather than one of our arms. Is that right? If there is a form of it we could run before sending a patch we would like to, and if not we will simply say in future PR bodies which chapter a new name is reached from, so the check is cheap for you to run on your side.

The two things we got wrong on this PR — the `Core/` path and the selective cite — both stand as written.


---

damiant3:
Shipped in Update 56: commit 6cd2ca1bd9dddca5e6358021bbc0fb5d6d5e874e on GitHub master and GitLab main (2026-09-08), carrying the landing recorded above (Perforce main 22475). Credit is in the release note, docs/PM/Active/GitHubUpdates/GitHubUpdate56.md, paragraph "Outside contributions". Thank you.

PR #118, opened 2026-09-03
zig plug: the five memory builtins Update 54 reaches, and six Nothing-returning fragments that returned i64

showell:
Backlog row: `codex/plugs/plugs-backlog.md` **2.19** (added in this PR). Zig plug only; no other plug is touched.

## Why

**Five builtins had no emitter entry**, so any program reaching them refused: `peek-32`, `poke-32`, `alloc-bytes`, `poke-byte`, `__memset`.

Update 54's `check-batch-*` memo table reaches all five — the slot table is `alloc-bytes`, the key and value cells are `peek-32`/`poke-32`, and the poison is `__memset`. Before this the zig plug could not transpile the compiler at all.

**And six fragments that were already present emitted the wrong return type.** These builtins carry `NothingTy` in `Types/Builtins.codex`, and `ZigEmitter` maps `Nothing` to `void` (`NothingTy -> "void"`), but the fragments emitted `) i64 { ... return 0; }`:

```
cx_heap_advance   cx_deck_enter   cx_deck_set
cx_heap_restore   cx_deck_exit    cx_memset
```

That was invisible for as long as every call site emitted `_ = f(...)`, which is legal either way. It becomes a build error the moment a `-> Nothing` definition's whole *body* is such a call — which `check-batch-poison` is:

```
check-batch-poison (base) (byte) (len) =
 if byte == 0 | len <= 0 then __heap-advance 0
 else __memset base byte len
```

`cx_print` and `cx_print_line` were already `void`, so the convention existed and these six were out of step with it.

**`cx_address_of` gained an enum prong.** A payload-free Codex union is a bare zig enum, and bare metal hands back the tag word — the same identity the switch already gives an Integer. U54's `mcopy-real-width` is the first caller. Its `else` now uses `@compileError(... ++ @typeName(...))`, so an unhandled type names itself rather than saying "no address-of for this type" three instantiations deep, which is how the previous message was actually encountered.

## What is measured, and what is not

**Measured.** `native/codexir` — the whole compiler, 2.7 MB of source, through the seed under QEMU, then the plug, then `zig build-exe` — builds and runs. It could not before this: transpilation refused on the missing builtins, and after they were added the build failed on `expected type 'void', found 'i64'` and three `no address-of for this type`, which is how the second and third parts of this PR were found.

**Not measured, and I would rather say so than imply otherwise:**

- I have not run your gate.
- No non-zig plug was touched or run. The `-> Nothing` mismatch is a zig-specific consequence of `NothingTy -> "void"`; whether any other text plug has the same shape is unchecked.
- The five builtins are exercised by the paths U54's check compact takes and by the compiler build. I have not written a targeted corpus program per builtin.
- The enum prong is exercised by one caller (`mcopy-real-width`). The identity it returns matches bare metal's by reasoning about the tag word, not by a differential measurement against a bare-metal run of that function.


showell:
Two more commits pushed onto this branch, both in its subject area.

**`zig plug: poke-16, port-out-byte and __self-type-defs`** — the rest of the refusal set. Update 55 made `Chapter: Opening` bundlable, and a subject that carries the driver reaches these three; C# and wasm emit all three, zig emitted none. `poke-16` is `cx_poke_16` shaped like the `cx_poke_32` this branch already adds; `port-out-byte` takes both arguments and answers 0, as the C# plug does, since a hosted process has no I/O ports; `__self-type-defs` is nullary — its registry type is not a function — so it goes in the name table rather than among the argument-taking emitters.

Emitted and RUN at this base, not just read:

```
probe (b) = act
 p <- port-out-byte 1016 65
 let w = poke-16 b 0 4660
 in let back = bit-and (peek-32 b 0) 65535
 in back + p
```

4660 is 0x1234, so a correct 2-byte little-endian store reads back as 4660 and a port write answering 0 leaves it alone. The transpiled program prints `4660` — which exercises this branch's own `peek-32` and `alloc-bytes` at the same time.

`__self-type-defs` is NOT exercised by that probe: only the compiler uses it, and the emitted zig contains no `cx_ll_empty(TypeBinding)`. Stated rather than implied.

**`zig plug: run.ps1 creates its output directory and falls back to the seed`** — found by needing it. In a fresh checkout the script asked compile.ps1 to write a log into a directory nothing had created, so the failure surfaced as `IR compile failed; see <that log>`, naming a file it had just failed to create; and it then refused for want of a self-hosted kernel that a plug run does not need. Both are why the probe above could not run until they were fixed.

damiant3:
Landed, with credit. **PR 118.**

Two changes from the patch as submitted, both flagged in the backlog row:

- The new fragment block put `zig-p-cx-poke-32`'s signature above `zig-p-cx-poke-16`'s body, and left `cx_poke_32` and `cx_port_out_byte` without one. All 98 other `zig-p-*` definitions in the file are adjacent to their own signature, so each new one is paired here. You fix this yourself later in the branch (cc6eab7e).
- `cx_poke_16` and `cx_port_out_byte` were missing from `zig-prelude-decls`. A prelude name absent from that list is one source identifier away from exactly the collision PR 119 fixes.

The `run.ps1` half here was superseded by PR 130's version, which passes `-Kernel` in both branches.

showell:
Thank you for landing it, and thank you for fixing the two things rather than bouncing the patch.

**How can we do better here?** Both of your changes are mechanical, both are conventions of the file that we could have read off the file itself, and we then repeated one of them in PR 128 (`cx_ll_copy`). Three names across two PRs is a habit, not a slip.

What we missed:

1. **A `zig-p-*` definition is adjacent to its own signature.** All 98 existing ones are; our new block put `zig-p-cx-poke-32`'s signature above `zig-p-cx-poke-16`'s body and left two fragments with no signature at all. We fixed it ourselves later in the branch (cc6eab7e) without noticing it was a rule we had broken rather than a typo we had made.
2. **A prelude name must appear in `zig-prelude-decls`.** Your note that an absent name is one source identifier away from exactly the collision PR 119 fixes is the part we did not understand — we had read that list as documentation rather than as the reservation set.

So, concretely:

- **Is there a linter for either of these that we could run before sending?** We would rather run yours than write our own and have it disagree with you at the edges. If there is not one, we are willing to write both — a signature-adjacency check over `zig-p-*` and a cross-check that every `cx_*` a fragment defines appears in `zig-prelude-decls` — and send them as a PR for you to keep or discard. Say the word either way; we do not want to add a tool to a tree that does not want it, which is the same question PR 129 raises about `validate-naga.mjs`.
- **Are there other lists of this shape in the zig plug?** `zig-prelude-decls` is the one that bit us twice. If there are two or three more registries where adding an emitter means adding a row somewhere else, knowing the set now is cheaper for both of us than finding them one review at a time.

We look forward to testing the arc when Update 56 reaches us. The five-PRs-as-one-arc decision makes sense from here, and it is useful to know that is how a related batch will land — we will keep sending them separately, since that is what keeps each diagnosis attached to its own evidence, but we will say in the body when we know one depends on another.


---

showell:
**Withdrawing the question above — the linter exists and so does the written rule, and we ran neither.**

`build/check-zig-prelude-surface.ps1` asks exactly the two questions we got wrong:

> It reconstructs the whole from this table and asks the two questions a new part can get wrong: is every name in the emitted surface covered by `zig-prelude-decls`, and is every emitted prelude a sub-selection of the whole in table order. Run it after changing the prelude.

And the rule is written down immediately above the table we edited, in `ZigEmitter.codex` itself:

> THIS TABLE IS THE SOURCE. A NEW PART IS WRITTEN HERE BY HAND: a `zig-p-` fragment list whose own declaration site is a `ShakeUse` of its name, a `ShakePart` row registering it below, and its name added to `zig-prelude-decls`.

Three things. We did two of them, three times, across two PRs. The instruction was in the file we were editing and the checker was in `build/` next to twenty-eight others, and we asked you whether either existed.

So: no linter needed from us, and please disregard the offer to write one. What we owe you instead is a habit, and it is the same one PR 132 taught from the other direction — **read what the tree says about the table before editing the table.** Its own prose is a specification and we treated it as commentary.

Two things still worth asking, both smaller than the question we withdrew:

1. **Can we run it?** It takes `-Subjects` and reconstructs from emitted output, so it may need the real compile path rather than one of our arms. If it is runnable outside your environment we will run it before sending anything that touches the prelude. If not, we will say in the PR body that the prelude surface changed, so it is cheap for you to know when to run it.
2. **`check-plug-builtins.ps1`, `check-plug-guards.ps1`, `check-plug-types.ps1`, `check-plug-ports.ps1`, `check-plug-field-slot.ps1`** — is there a short list of which of these a plug PR should have passed, the way this one should have passed the prelude surface check? We would rather learn the set once than discover it one review at a time. This is the same question as before, asked properly: not "does a tool exist" but "which of your tools does this kind of change owe you".


---

damiant3:

PR #119, opened 2026-09-03
zig plug: a closure parameter named p0 collides with a source p0, so name them in the emitter's namespace

showell:
Backlog row: `codex/plugs/plugs-backlog.md` **2.20** (added in this PR). Zig plug only. Independent of #118, though both were found by the same build.

## Why

`zig-closure-pass` and `zig-closure-params` named a lambda's parameters `p0`, `p1`, ... Zig forbids shadowing, so a Codex definition holding a local of the same name emits invalid zig:

```
function parameter 'p0' shadows local constant from outer scope
```

**The emitter could not see it coming**, because the colliding name is one the *source* chose.

Update 54 made it live. `check-chapter`'s CHECK-REG instrumentation binds `p0`..`p5`:

```
"CHECK-REG tdm=" & show (p1 - p0) & ...
```

and `check-chapter` also builds its slug cache with a comprehension, which is a closure. That put a `p0` parameter inside a scope already holding a `p0` constant.

## The fix

`_cp0`, `_cp1`, ... — putting closure parameters where `_Env`, `_ctx` and `__lam_` already live: the emitter's own namespace, which a Codex identifier cannot sanitize into.

Worth naming the general shape, since this is only one instance of it: **an emitter that names anything in the program's namespace is one source identifier away from a collision.** This PR closes the instance. Whether any other emitted name in `ZigEmitter` is still in the program's namespace is not something I audited — I fixed what the build hit.

## What is measured, and what is not

**Measured.** The compiler transpiles and `native/codexir` builds and runs. This was found by that build: after the missing builtins in #118 were added, transpilation got through for the first time and landed straight on this.

**Not measured.** I have not run your gate. No non-zig plug was touched. I have not constructed a minimal reproducer — the case that found it is the compiler's own `check-chapter`, which is a large subject; a two-line Codex program with a `p0` local and a comprehension should reproduce it, and I have not written one.

### DIFF
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -6662,6 +6662,28 @@ advanced: `apps/classics-test` moved off `undefined local variable $real_to_int`
 to a wrong answer. The fix state was hashed before the control ran and verified
 after restoring it.
 
+## 2.20 -- CONTRIBUTED by Steve Howell (PR pending, 2026-09-03): an emitted closure parameter was named `p0`, which zig refuses when the Codex source has its own `p0`
+
+`zig-closure-pass` and `zig-closure-params` named a lambda's parameters `p0`,
+`p1`, ... Zig forbids shadowing, so a Codex definition holding a local of the
+same name emitted invalid zig -- "function parameter 'p0' shadows local
+constant from outer scope" -- and nothing in the emitter could see it coming,
+because the colliding name is one the SOURCE chose.
+
+Update 54 made it live. `check-chapter`'s CHECK-REG instrumentation binds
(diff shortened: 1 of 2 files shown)


damiant3:
Landed, with credit. **PR 119.**

All three sites taken (`zig-closure-pass`, `zig-closure-params`, `zig-closure-pass-tail`). A census of `"p" & integer-to-text` finds no fourth, so the instance is closed.

showell:
Thank you — and thank you for running the `"p" & integer-to-text` census before accepting rather than taking our word that three was all of them. A patch that closes an instance is worth less than one that closes a class, and only your side could confirm which this was.

We look forward to testing it when Update 56 reaches us. Nothing to re-verify here until then — the work is in Perforce and we are reading your note, not your tree.


---

damiant3:

PR #127, opened 2026-09-05
mcopy-type reads its tag with variant-tag, not peek-qword word 0 (root of issue 126)

showell:
Branch `mcopy-type-variant-tag`, one commit, cherry-picked clean onto
`675a0775` (Update 55) with no stack under it.

**On the base, stated narrowly.** U55's `mcopy-type` is identical to the
function measured, and the cherry-pick was clean. That is all we claim about
the base: **the measurements below were taken on our own branch at U56, with
our four-page zig emitter split under them, and have NOT been re-taken at
U55.** Nor has the bare-metal proof battery been run on this branch — the
argument that bare metal is unaffected is from the emitters, below, not from
a measurement.

`Ladder: mcopy-tag-read-126`

---

## What it changes

`mcopy-type` validates a box by reading a tag out of word 0:

```
   else if peek-qword a 0 < 0 then ErrorTy
   else if peek-qword a 0 > 28 then ErrorTy
```

28 is the `CodexType` variant count, so this is a range check on a tag. Word
0 **is** the tag on bare metal, where a variant's first word is its tag. It
is a **payload** word on any target whose unions lay themselves out
differently. The zig plug's read comes back a pointer, fails `> 28`, and
every type this touches escapes to `ErrorTy`.

The escape is silent: `check-errors` stays 0, so nothing reports it. The
compact then copies nothing and the originals are reclaimed behind it, so the
damage surfaces later and somewhere else.

The fix asks the question with the builtin that answers it:

```
   else if variant-tag ty < 0 then ErrorTy
   else if variant-tag ty > 28 then ErrorTy
```

## Why bare metal is unaffected

`emit-variant-tag-builtin` (`Emit/X86_64Builtins.codex:170`) emits
`mov-load rd, r, 0` — a 64-bit load from offset 0 of the value. `address-of`
is `emit-identity-builtin`, so `peek-qword a 0` is the same 64-bit load from
the same address. The two are the same instruction on this target.

The dereference needs `ty` to be non-null and in the copied region, and the
two guards immediately above (`a == 0`, `a < mc.mc-floor`) already establish
both before either branch is reached.

## Precedent

`Emit/X86_64Compound.codex` moved off this exact idiom already. Its comment
says it used to read the tag with `peek-qword (address-of ty) 0`, "which is
the same load on bare metal but silently answers 0 on any target where
`address-of` cannot be modelled: that cost the C# arm every tag in this
table." That fix was applied to the walker. This is the same mistake in the
copier.

## What it is measured to fix

On the zig arm at U56, with a harness that calls `compile-frontend-cdx`:

| | before | after |
|---|---|---|
| escapes on a 7-line subject | 3 × `FunTy -> ErrorTy` | 0 |
| every definition's type | `error` | correct |
| `fib` transpiled | garbage — `@compileError("no zig type")` in the signature | **byte-identical** to the known-good `fib.zig` (14,877 bytes) |
| a 388 KB subject | crash | 556,955 bytes, 690 functions |

With one further zig-plug change (not in this PR — it depends on our emitter
split), the transpiler's fixed point holds: pass 1 under QEMU and pass 2 as a
native binary both 2,932,307 bytes, byte-identical, and the sample program
matches its expected output on all 9 lines.

## Relationship to issue 126

Issue 126 reported that the plug-built compiler types every comparison
`error` where bare metal says `boolean`, and we assessed it as inert because
`fib`'s single comparison did not change the emitted zig. **This is its root,
and it is not inert.** It is not one comparison — it is every type the copier
touches, and the reason it looked inert is that the copy was being discarded
anyway on the arm we measured.

## What we have not checked

Whether the C# and wasm arms improve. We have not run them.

### DIFF
--- a/codex/compiler/Types/Unifier.codex
+++ b/codex/compiler/Types/Unifier.codex
@@ -1384,6 +1384,16 @@ Section: Deep Type Copy
    else let c = list-at cs i
    in mkey-ctors-loop cs (i + 1) len depth mc (cons-mix (cons-mix (cons-mix acc (address-of (mcopy-name (c.name) mc))) (address-of (mcopy-types (c.fields) depth mc))) (address-of (mcopy-types (c.return-type) depth mc)))
 
+ THE BOX CHECK READS A TAG, so it asks for one. `peek-qword a 0` is that tag
+ on bare metal, where a variant's first word IS its tag -- and it is a payload
+ word on any target whose unions lay themselves out differently. The zig plug's
+ read came back a pointer, failed `> 28`, and every type this touched escaped
+ to ErrorTy with the bag still empty: the compact then copied nothing and the
+ originals were reclaimed behind it. `variant-tag` is the builtin for this
+ question and it costs the same load on bare metal;
+ `Emit/X86_64Compound.codex` moved off the same idiom for the same reason
+ after it "cost the C# arm every tag in this table".
+
   mcopy-type : CodexType, Integer, MemoCtx -> CodexType
   mcopy-type (ty) (depth) (mc) =
    let a = address-of ty
@@ -1391,8 +1401,8 @@ Section: Deep Type Copy
    else if a < mc.mc-floor then ty
    else if __heap-save >= mc.mc-ceiling then ErrorTy
    else if depth >= max-recursion-depth then ErrorTy
-   else if peek-qword a 0 < 0 then ErrorTy
-   else if peek-qword a 0 > 28 then ErrorTy
+   else if variant-tag ty < 0 then ErrorTy
+   else if variant-tag ty > 28 then ErrorTy
    else mcopy-type-memo ty a (memo-probe a mc) depth mc
 
   mcopy-type-memo : CodexType, Integer, Integer, Integer, MemoCtx -> CodexType


damiant3:
Landed, with credit. **PR 127.**

Ingested into the Perforce depot as blu 22432 (source) and 22433 (seed), on main as **22435**. Seed `94A63C59FD218528D7C476CB86C8ABAE2D5145F7E45E18527070DA0ACC55A899`.

Your two lines went in as written. The prose above them was trimmed to the rule rather than the account -- this tree has a standing rule against narrative in source -- so it keeps why `peek-qword a 0` is a bare-metal-only contract and that `variant-tag` costs the same load there.

I ran the census your change invites before accepting it: no sibling of this shape survives in target-independent code. The other `peek-qword <x> 0` sites are memo-slot keys in the same file, diagnostics, and `Emit/X86_64Compound.codex`, where assuming the bare-metal layout is correct.

Proof, each a separate run:

- scratch fixed point, **one pass**: stage2 == stage3, 3,179,918 bytes
- BVT on that candidate: **137 pass, 0 fail**
- signed, then self-verified -- and the verifier was compiled by the *previous* seed rather than by the candidate, so the artifact under test did not build its own judge

The compiler came out **16 bytes smaller**, and that delta is attributable to this change: the previous seed was current with its source, and signing patches in place without altering length.

Thank you. This one was small and load-bearing, and the reasoning about which target the representation contract actually belongs to is what made it easy to accept.

showell:
Thank you — and thank you for running the `peek-qword <x> 0` census before accepting. Knowing that the surviving sites are memo-slot keys, diagnostics and `Emit/X86_64Compound.codex`, where the bare-metal layout is the correct assumption, is worth more to us than the fix: it tells us the shape of the rule rather than the instance.

**The trimmed prose is a constraint we did not know about, and we would like to understand it properly, because it touches every PR we send.**

> The prose above them was trimmed to the rule rather than the account -- this tree has a standing rule against narrative in source.

We take that as given and will write to it. What we are unsure of is where the line falls, and getting it wrong wastes your reviewers' time on every future patch. Concretely:

- **Rule vs account.** Our habit is to write *why* a line is the way it is, often including what the wrong version did — "`peek-qword a 0` is a bare-metal-only contract; `variant-tag` costs the same load there" is the rule, and "this is the root of issue 126, found by X" is the account. Is that the split? Or is any explanatory prose in source unwelcome, with the *why* belonging in the changelist?
- **Is there a written style rule we can read?** If one exists in the tree we will follow it directly rather than infer it from what survives review. We could not find one; we may have looked in the wrong place.
- **Does it vary by file?** Some chapters carry long prose blocks that read as design documents — `Types/Unifier.codex` on the memo table, for one. If those are grandfathered, or if a chapter-level prose block is a different thing from a comment above a definition, that would explain most of what confused us.

**And a related one, since it is the same question from the other side.** Our PR bodies are long. That is deliberate — we cannot run your gates and long prose is how we hand you the evidence instead — but if the *bodies* are also more than you want, we would rather be told than keep spending your reading time. You have called the reasoning reviewable twice in this batch, so we have assumed not, but it is worth asking directly.


---

damiant3:
Shipped in Update 56: commit 6cd2ca1bd9dddca5e6358021bbc0fb5d6d5e874e on GitHub master and GitLab main (2026-09-08), carrying the landing recorded above (Perforce main 22435). Credit is in the release note, docs/PM/Active/GitHubUpdates/GitHubUpdate56.md, paragraph "Outside contributions". Thank you.

PR #129, opened 2026-09-05
wgsl plug: emit shaders Firefox accepts, all three causes — 81 of 81 pass naga, was 57

showell:
Branch `wgsl-firefox-naga`, three commits, cut from `675a0775` (Update 55)
with no stack under it. All three causes, not two. Chrome uses Tint, Firefox uses naga, and **Tint is the
permissive one** — a shader tested only in Chrome is untested rather than
correct.

**On the base.** Both commits apply to U55 and the gate was run there, not
only on our own branch. The plug consumes IR from the **seed** compiler
(`run.ps1` pins it: `seed/Codex.cdx [81F9E8171DCF6268]`), so the frontend
behaviour under this measurement is U55's — the same as the base.

`Ladder: wgsl-firefox-naga`

---

## Measured at this base, with the plug actually run

| population | before | after |
|---|---|---|
| the 24 gpushow kernels that failed | 0 pass | **24 pass** |
| all 42 gpushow kernels | 18 pass | **42 pass** |
| the whole gate population | 57 of 81 | **81 of 81** |

The 42 regenerated shaders are included. They are generated artifacts and
leaving them stale would leave the gate red after a merge.

## Cause 1 — a tail-call loop leaves the function with no terminator

25 files, 31 functions. Every path inside the loop returns or continues, so
no `break` is emitted; Tint's reachability sees the end is unreachable, and
naga supplies an implicit `return;` that disagrees with the declared type.
`wgsl-emit-loop-helper` now closes with an unreachable return of the zero
value, and `wgsl-zero-text` mirrors `wgsl-ty-text`.

## Cause 2 — a storage pointer cannot be a function parameter

WGSL restricts pointer parameters to the function, private and workgroup
address spaces; `storage` needs `unrestricted_pointer_parameters`, which
Tint implements and naga does not. **The pointer bought nothing**: the
buffers are module-scope `var<storage>` and the entry point already reads
them by name — only helpers took the long way round.

Dropping the parameter is not enough, and the second commit is the rest of
it. A helper names a buffer by **its own** parameter name while the global is
named for the **kernel's**: `ss-gather (buf)` is called `ss-gather gdep`, and
the global is `ssao_step_gdep_buf`. So the emitter needs the caller's
argument, not the callee's parameter, and it needs to know which kernel the
helper serves.

Both are recorded at the call site, where the caller, the callee's parameter
and the argument are all in hand; resolution then walks callee to caller
until it reaches a kernel. **A self-call is not a binding** — a self-recursive
helper passes its own parameter through unchanged, and is usually defined
before the kernel that calls it, so that useless bind would otherwise be the
one found first.

This rests on a property checked over the whole tree: **no helper is called
with more than one kernel's buffers.** A helper reached from two kernels
would need one emission per kernel; there are none, and a second would be
worth a refusal rather than a silently wrong name.

## Cause 3 — a bitcast is not a constant expression

A Real top-level constant emits a module-scope `const`, whose initialiser must
be a constant expression:

```
const pb_rad : f32 = bitcast<f32>(1057300152u);
error: Not implemented as constant expression: bitcast built-in function
```

naga does not implement `bitcast` there and Tint does, so Chrome accepted all
of these and Firefox refused four kernels outright. `0x1.0a3d70p-1f` IS the
f32 whose bits are 1057300152, written out — exact by construction, no
rounding, and it needs only the integer arithmetic already in the chapter.
The `f` suffix pins the type rather than leaning on AbstractFloat conversion.
Verified over 428,984 sampled f32 bit patterns, every one round-tripping
exactly; denormals, negative zero and `p+0` were each checked against naga
directly. An f64 too large for f32 now REFUSES rather than emitting a finite
stand-in, because a wrong number in a shader that compiles is worse.

30 of the 42 regenerated shaders carry hex float literals as a result.

## Reproducibility

The committed shaders are what this emitter produces. Built at this base and
regenerated, `SsaoKernel`, `PbrKernel` and `ReflectKernel` come out
byte-identical to the files in the diff.

## Also here

`apps/gpushow/tools/validate-naga.mjs`, the gate — see its own prose for why
it exists beside `validate-all.mjs`, which is careful about the population
and silent about the oracle.


showell:
Eye-tested in a browser, beyond the static gate: `ssao.html` and `reflect.html` both render from the regenerated shaders, served over an SSH tunnel so the page has the secure context WebGPU requires.

Worth stating because naga only validates -- it does not run the shader. These two are backed by `SsaoKernel` and `ReflectKernel`, both of which were in the failing 24 and both of which exercise cause 2: their helpers now read the module-scope buffer by name (`ssao_step_gdep_buf[idx]`) instead of dereferencing a storage pointer parameter.

showell:
Correction to this PR, made before review: it now carries THREE commits, not two.

The shaders in the original diff were regenerated on a tree that already had the cause-3 fix, so 30 of the 42 carried hex float literals that the emitter in the PR would not have produced — the artifacts could not be reproduced from the emitter shipped beside them. `cause 3: a hex float literal is a constant expression and a bitcast is not` is now included, which is where it belonged.

The body previously said cause 3 was "no longer observed" because naga accepts `bitcast` in a constant expression. That was wrong: it was not observed because it had already been fixed. naga does not implement it, which is exactly why Firefox refused four kernels that Chrome accepted.

Checked rather than assumed: built at this base, `SsaoKernel`, `PbrKernel` and `ReflectKernel` regenerate byte-identical to the committed files, and the gate still reports 81 of 81.

damiant3:
Landed, with credit. **PR 129.**

Ingested into the Perforce depot as blu 22421, on main as 22423.

Your `WgslEmitter.codex` was taken **exactly**: our copy was byte-identical to your PR base once line endings were normalised, so your head revision went in verbatim rather than hand-applied, then written with this tree's CRLF endings. The prose stayed as you wrote it -- it documents naga and Tint behaviour, which is code we do not own.

The 42 kernels were **regenerated from source through the rebuilt plug**, not copied from the PR, and then compared against yours:

- MOVED 35, and they are exactly the 35 this PR changes
- UNCHANGED 7, the ones it does not touch, byte-identical to our depot
- REFUSED 0
- and all 42 regenerated files are byte-identical to the same file on your branch

Two independently built emitters agreeing on 42 artifacts is a stronger check than either alone, and the 7 untouched kernels are the control saying the run moved only what the change moves.

One thing held back, and it is not a judgement on the tool: `tools/validate-naga.mjs` is not ingested. It drives `naga`, a Rust binary, from node, and this tree has a standing rule against adding dependencies outside its own Windows + codex-vm environment. That is the owner's call and it is open with him, not refused. The grading above did not need it.

Thank you -- the three causes were separated cleanly and the hex-float argument in particular made that change easy to accept.

showell:
Thank you — and the regeneration check is a better result than anything we sent. Forty-two kernels rebuilt from source through your independently built plug, 35 moved and they are exactly the 35 this PR changes, 7 unchanged as the control, all 42 byte-identical to our branch. Two emitters that share no build agreeing on 42 artifacts says something neither of us could say alone, and we will steal the shape: **regenerate from source through the rebuilt tool, and keep an untouched set as the control that says the run moved only what the change moves.**

**On the tool held back — understood, and we want to understand the rule rather than argue the case.**

> `tools/validate-naga.mjs` is not ingested. It drives `naga`, a Rust binary, from node, and this tree has a standing rule against adding dependencies outside its own Windows + codex-vm environment.

That is a coherent rule and a good one; a tree that self-hosts loses something real the first time it needs a toolchain it did not build. We will keep the validator on our side and keep grading with it, and we will keep reporting the number in PR bodies as an outside measurement rather than proposing the tool again.

The question we would like answered is about the general case, not this file:

1. **Is there a shape an external validator COULD take that would be ingestable?** For instance: a Codex program that reads a WGSL file and checks the subset your emitter targets, needing no node and no Rust — slower and less authoritative than naga, but inside the rule. Or is the answer that outside validation belongs outside the tree by design, and the right artifact is the *number* in a PR body rather than the tool that produced it?
2. **Is `naga` specifically interesting to you as a reference even though the driver is not?** It is the validator Firefox actually uses — Chrome's Tint is the permissive one, which is why four kernels passed there and failed in Firefox — so "what does naga say" is a question about a real target and not about our tooling. If that answer is useful, we can keep producing it on request without anything entering your tree.

Both questions apply beyond WGSL. Most of what we can offer is measurement against implementations you do not own, so knowing which of that is welcome as a number, which as a tool, and which not at all would save us both time.


---

damiant3:
Shipped in Update 56: commit 6cd2ca1bd9dddca5e6358021bbc0fb5d6d5e874e on GitHub master and GitLab main (2026-09-08), carrying the landing recorded above (Perforce main 22423). Credit is in the release note, docs/PM/Active/GitHubUpdates/GitHubUpdate56.md, paragraph "Outside contributions". Thank you.

PR #130, opened 2026-09-05
zig plug: emit real-to-int and real-from-int (replaces #100, no cross-graded expectation)

showell:
Replaces #100, which is closed. One commit, cut from `675a0775` (Update 55)
with no stack under it. **The test that blocked #100 is not here** — the
question it raised is settled upstream.

`Ladder: zig-real-int-conversions`

---

## What the review said, and what changed

#100's review found the emitter correct: NaN caught by `v != v`, both
infinities by the range tests, the top bound exact, `-2^63` admitted by `<`
rather than `<=`, `-0.0` truncating to 0 the way `cvttsd2si` does, and no
input where the zig arm disagrees with the x86 one.

**The blocker was the expectation, not the emitter.** #100 put NaN, infinity
and overflow rows in `codex/test/ops`, which `build/test-cross-batch.ps1`
grades on arm64 and riscv64 as well — and those three cases are exactly where
the ISAs disagree: x86 answers the integer indefinite, arm64 and riscv
saturate.

That is settled now, by you: `real-to-int-wide.codex` carries the unambiguous
range for all three backends, and says why in its own prose. **So this PR adds
no cross-graded expectation at all** — the emitter alone, and the two script
fixes the review asked for.

## Measured, transpiled and run

The plug emits `real-to-bits` and `bits-to-real` and neither of the two
conversions beside them, so a program that turns a Real into an Integer or
back refuses at the emitter. `show` on a Real routes through the same
conversion, so it is not an exotic path.

| probe | answer |
|---|---|
| `real-from-int` then `real-to-int` of `3000000000` | `3000000000` |
| the same of `-3000000000` | `-3000000000` |
| `real-to-int 2.75` | `2` |
| `real-to-int -2.75` | `-2` |

3000000000 is above 2^31 on purpose — the range a 32-bit conversion saturates
in, which is where the riscv encoder defect #100 exposed lived. Truncation is
toward zero rather than toward the floor. Only unambiguous values, the
standard `real-to-int-wide` sets.

## The two script notes from the review

`run.ps1` creates the `build-output` directory it writes its log into. Without
it a fresh checkout failed as `IR compile failed; see <that log>`, naming a
file it had just been unable to create.

And it passes `-Kernel` in **both** branches, which is the half the earlier
version got wrong: testing an absolute path and then passing nothing left
`compile.ps1` on its own relative default, so from any working directory but
the repo root the test passed and the compile failed anyway. Passing it
explicitly also puts the choice in the `kernel:` digest line. This is related
to backlog 2.26 — the same scripts sharing a fixed scratch path — and is not
that defect.

## Backlog

Row **2.29**, since 2.06 is taken on your side.

### DIFF
--- a/codex/plugs/plugs-backlog.md
+++ b/codex/plugs/plugs-backlog.md
@@ -7500,6 +7500,33 @@ plug fingerprint beside it already guards the EMITTER for this reason; this guar
 the SOURCE. Calibrated both ways: the stale concat exits 2 with the refusal, the
 regenerated one passes.
 
+## 2.29 -- OPEN (Steve Howell, 2026-09-05): the zig plug cannot emit real-to-int or real-from-int, and run.ps1 chose a kernel it did not pass
+
+The plug emits `real-to-bits` and `bits-to-real` and neither of the two
+CONVERSIONS beside them. A program that turns a Real into an Integer, or back,
+refuses at the emitter -- and `show` on a Real routes through the same
+conversion, so it is not an exotic path.
+
+The emitter is the increment; the EXPECTATION is what kept the earlier attempt
+out, and that part is settled now. PR 100 put NaN, infinity and overflow rows
+in `codex/test/ops`, which `build/test-cross-batch.ps1` also grades on arm64
+and riscv64, and those three cases are exactly where the ISAs disagree: x86
+answers the integer indefinite, arm64 and riscv saturate. `real-to-int-wide`
+now carries the unambiguous range for all three backends, so nothing here adds
+a cross-graded expectation at all.
+
+Measured, transpiled and RUN rather than read: `real-from-int` then
+`real-to-int` round-trips 3000000000 and -3000000000 -- above 2^31 on purpose,
+the range a 32-bit conversion saturates in -- and `real-to-int` truncates 2.75
+to 2 and -2.75 to -2, toward zero rather than toward the floor.
+
+Also here: `run.ps1` creates the `build-output` directory it writes its log
+into, and passes `-Kernel` in BOTH branches. Testing an ABSOLUTE path and then
+passing nothing left `compile.ps1` on its own RELATIVE default, so from any
+working directory but the repo root the test passed and the compile failed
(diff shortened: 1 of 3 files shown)


damiant3:
Landed, with credit. **PR 130.**

This `run.ps1` supersedes PR 118's one-branch form, for the reason you give: testing an absolute path and then passing nothing leaves `compile.ps1` on its own relative default. It also closes the zig instance of a finding raised here the same week -- 38 of our 49 plug `run.ps1` pass no `-Kernel` at all -- leaving 37.

One gap found while grading, recorded open rather than fixed: `real-approx-to-int` has no emitter either, so `codex/test/ops/real-to-int-wide` transpiles but does not build, carrying `@compileError("zig plug: no emitter for real-approx-to-int")` on the `approx-wide` line. The refusal self-identifies, which is the right shape.

showell:
Thank you — and thank you for taking the `run.ps1` from this one over 118's. The one-branch form was our bug: testing an absolute path and then passing nothing leaves `compile.ps1` on its own relative default, which is the failure that is hardest to see because it looks like the test passed.

**On the gap you recorded open:**

> `real-approx-to-int` has no emitter either, so `codex/test/ops/real-to-int-wide` transpiles but does not build, carrying `@compileError("zig plug: no emitter for real-approx-to-int")` on the `approx-wide` line.

We would like to take that one. Before we write it, two questions so we send the right thing rather than a plausible one:

1. **What are the semantics?** `real-to-int` is a trapping conversion. Is `real-approx-to-int` the same conversion over an f32, or the same width with a different rounding or overflow contract? The name reads as the `approximate` qualifier — `RealTy RwF32` — but we would rather ask than infer, since a conversion that is wrong at the boundary is exactly the kind of defect that never shows up in a passing test.
2. **Is there a reference we should be grading against?** The C# plug has been the best crib for us; if it emits this one, that is what we will match, and if it does not, we would rather say so in the PR than pick a behaviour.

That `@compileError` carrying the builtin's own name is a good shape and we will keep to it for anything we cannot emit.


---

damiant3:

PR #131, opened 2026-09-06
zig plug: address-of answers 0 for every source literal, and the checker keys type identity on it

showell:
`cx_address_of` is heap-relative, and its last line is

    return if (cx_p >= cx_base) @intCast(cx_p - cx_base) else 0;

A string literal lives in `.rodata`, below the heap base, so it takes the
`else`. **Zero is not a spare value there**: the zero-length-slice case two
lines above uses it deliberately to mean "no pointer, share as-is", and the
compiler reads it as an answer.

Six lines, built by the plug's own binary:

```
literal 0  other 0  computed 6291456  empty 0
```

Two distinct non-empty literals, one address between them.

## What reads it

`mcopy-name-fresh` keys a `Name` on `cons-norm (cons-mix 701 (address-of tv))`
and nothing else, so every literal-named `Name` collides on one key and the
first one copied is adopted by all the rest. The mechanism is not confined to
names — the content keys are all built this way, and
`is TypeCon (n) -> cons-mix 26 (address-of (mcopy-name n mc))` keys a type on
its name's canonical address alone.

## The observable

Every effect label in a compiled program becomes the same one. One program,
six labels:

| arm | labels |
|---|---|
| bare metal @ 53b3b213 | `Device.Mmio`×3, `Device.Port`×2, `Console.Write`×1 |
| bare metal @ U56 | the same |
| an independent Rust interpreter @ U56 | the same |
| **the zig plug @ U56** | **`Device.Mmio`×6** |

A six-line program using only `print-line-uni` reports its row as `Task` — a
name from the builtin table it never mentions.

## The fix

One line: return the signed displacement, so `.rodata` answers a distinct
negative. Every reader then gets what it wants — `a == 0` stays false so
identity survives, and `a < mc.mc-floor` and `address-of t < b` stay true,
which is the right answer because a literal never moves. Bare metal has the
same shape from the other side: its literals sit in the image below the heap
and its identity-`address-of` answers them distinctly too. Offset 0 cannot
collide with the empty-slice sentinel because `cx_hp` starts at 6291456.

**The empty literal moves off 0 as well and that was not the intent.** zig
types `""` as `*const [0:0]u8`, a pointer rather than a slice, so the
zero-length guard never fired for it either way. Behaviour is unchanged —
`a < mc.mc-floor` shares it exactly as `a == 0` did — and a runtime empty
slice, whose `.ptr` really is undefined, still hits the guard and still
answers 0.

## Evidence

- **Three bootstrap rounds.** The emitter emits the same bytes for its own
  source whether built from the old prelude or the new: `r3.zig == r2.zig`
  byte for byte. So the collision never reached emitted code, and the fixed
  point holds with the fix in. That also answers the hazard the fix
  introduces — literal addresses now vary with ASLR, and emission is stable
  regardless.
- **35 corpus programs** compiled to IR and compared against an independent
  arm: 35 byte-identical after, 0 before, and **nothing changed but the
  labels**.
- **safari's 54 specs** on the fixed binary, with the same three issue-125
  gaps it had before and no new ones.

**Not proven: that type identity was safe.** Zero of 35 programs changed
anywhere but their labels, and none of them had two literal-named builtin
types to tell apart. Absence in a sample is not confinement, and I would read
this as a wire defect with a type-identity mechanism rather than as bounded.

## Why it was invisible

`mcopy-name` runs only on names inside the copied region, and names from user
source are computed texts with real addresses. Only names built from literals
in the **compiler's own source** take the zero path. Both existing arms
consume the IR rather than read it, so a field neither looks at could be wrong
indefinitely; this surfaced on the first day of an arm that reads the wire.

The comment above `cx_address_of` describes this exact failure as Finding 31 —
*"answering a constant 0 made every object identical to every other one AND to
null, and the compiler reads that as an answer"*. This is the residue that fix
left behind, and #128 is the neighbouring case in the same function.

Backlog row 2.29. One plug (R-ONE); the compiler is not on the path, so no
token.

---

Reported and prepared by Claude (Anthropic), working with Steve Howell.


damiant3:
Landed, with credit. **PR 131.**

Taken as the arc's end state for `cx_address_of`, together with 118's enum prong and 128's empty-slice guard.

showell:
Thank you. Taking 118's enum prong, 128's empty-slice guard and this as one arc for `cx_address_of` is the right shape — they were three symptoms of one contract and we sent them as three PRs only because that is how we found them. We look forward to testing the arc end to end when Update 56 reaches us.

One thing worth stating plainly for whoever reads this later, since it is the reason this mattered at all rather than being a cosmetic fix: **the checker COMPUTES with `address-of`.** `mcopy-name-fresh` keys a `Name` by `cons-mix 701 (address-of tv)`, so a plug that answers 0 for every source literal does not produce a slightly wrong number — it makes every literal-named `Name` collide on one key, and the first one copied is adopted by all of them. It is the kind of defect that reads as an unrelated bug somewhere else entirely.


---

damiant3:

PR #132, opened 2026-09-06
scope: resolve each citation once per unit, not once per slug transition (a third of a self-compile)

showell:
`scope-adefs-ll` rebuilds a chapter's rename map at every slug transition of the
defs walk, and `apply-cite-overrides-loop` resolves every citation each time
through `find-slug-for-cite-name`. That function depends on the quire, the cite
name, the assignments and the slug cache — **and none of those changes while
the walk runs**, so it recomputes identical answers at every chapter boundary.

The product is *(slug transitions) × (citations) × (slug cache)*, and the cache
is one entry per **definition** rather than per chapter, which is what makes it
large. Compiling the compiler: 91 transitions, 106 citations, 6,431
assignments — **9,646 resolutions**, each walking 6,431 entries with a
character-by-character suffix compare per entry.

`scope-achapter` now resolves each citation once, beside the cache that feeds
them, and carries the list down; `apply-cite-overrides-loop` reads
`list-at cite-slugs i`.

## Measured

The compiler compiling itself, through a tree-walking interpreter where the
work is **counted rather than timed** — so the number is about the program and
not about the machine:

```
without   43,762,877,372 steps
with      29,306,453,914 steps      a third of the compile
```

and **the IR is byte-identical either way**: 19,570,302 bytes over 6,890
definitions, checked against an arm that shares no code with the one that
produced it.

Fourteen smaller subjects are byte-identical too, with their step counts
unchanged to a tenth of a percent. That is expected rather than disappointing:
a four-chapter bundle with one citation resolves four times and saves three. A
small subject cannot show this and should not be asked to.

## What this does not fix

The character loop underneath. A `text-contains` guard in front of it was tried
and **reverted** — this bundle *defines* `text-contains` as
`text-index-of s sub >= 0`, so the call site gets a Codex substring search
instead of the builtin and a self-compile got dramatically slower. CDX3005 says
exactly that will happen and is printed on every build of the subject; I had
filtered it as boilerplate, which is how I walked into it.

`codex/test/slug-suffix.codex` carries that finding along with the eleven pairs
a slug walk can present, including the one that discriminates a real
`ends-with` from a `contains`: `Foo--Bar--Baz` contains `--Bar` and does not
end with it.

## What would fix it

A lookup keyed by bare name would make the outer two factors constant. Nothing
inside `slug-has-suffix` can, and this change does not attempt it.

COMPILER-53. The compiler is on the path, so it takes the token.

---

Reported and prepared by Claude (Anthropic), working with Steve Howell.


showell:
A second, independent measurement of this change, on a different arm and in a different unit.

`codex-wasm-transpiler` builds a wasm module that compiles the source that produced it. Its subject is the compiler plus `codex/plugs/wasm/WasmEmitter.codex` — it does **not** contain `ZigEmitter`, so the only code difference from U56 in this build is this PR.

```
peak linear memory   2180.2 MB  ->  1655.1 MB      -24.1%
                     40.4% of wasm32's hard 4 GiB ceiling
fixed point          HOLDS
zig road vs self road: the two roads agree, byte for byte
```

Every number in the PR body is *steps* on a tree-walking interpreter. This one is **peak memory on a self-hosting wasm module**, and it moved by a quarter. The repository gates on that figure explicitly — it refuses a build more than 2% over its banked value, precisely because wasm32's 4 GiB is an address space rather than a policy — so the direction and size are both load-bearing there rather than incidental.

Worth saying plainly: I did not predict this. The change was reasoned about and measured as work avoided, not as memory saved. The allocation it removes is `find-slug-for-cite-name` rebuilding the same answers at every chapter boundary, and on a bump allocator that never frees, avoided work and avoided memory are the same thing.

---

Measured by Claude (Anthropic), working with Steve Howell.

damiant3:
Landed, with credit. **PR 132.** Perforce depot: blu 22497, on main as **22500**. Seed `39D40A84D3B90006F5F4B27188CF5DAAD183E83F9BDDA609E9D1AB31BE1DCAAE`.

Your `ChapterScoper` half went in exactly as written. Four call sites went in beside it, because the patch as submitted would have miscompiled.

**The blast radius.** `apply-cite-overrides`'s fifth parameter changes MEANING -- `slug-cache`, one entry per assignment (6,431 compiling the compiler), becomes `cite-slugs`, one resolved slug per citation (106). Both are `List Text`. The PR updates `ChapterScoper.codex` and leaves four callers passing the old list into a parameter now indexed by citation:

- `codex/compiler/IR/Lowering.codex` (twice)
- `codex/compiler/Semantics/NameResolver.codex`
- `codex/compiler/Types/TypeChecker.codex`

It compiles, and then produces wrong chapter renames in LOWER, NAME-RESOLVE and CHECK with no diagnostic. Each is now hoisted the same way you did it, at the point where `slug-cache` was already being built. Two of those (`lower-defs`, `lower-defs-acc`) turned out to be dead code -- fixed anyway, since a caller passing the wrong list is a defect whether or not anything reaches it.

**What proved it was an IR-equivalence arm, not the fixed point.** A mis-wired `cite-slugs` still compiles and still emits plausible output, so a green fixed point cannot tell a correct hoist from a wrong index. The compiler's own source emitted by the previous compiler and by the candidate came back byte-identical: 17,003,625 bytes, 6,890 definitions. Plus BVT 137/0, signed, self-verified.

**One thing I could not reproduce, said plainly because your title is about speed.** Emitting the compiler's IR takes 148.0s under the head compiler and 147.9s under this one -- 0.2s across 148s. An earlier reading of two 6-second self-compiles suggested 12%, and that was noise. Your measurement counted STEPS on a tree-walking interpreter, where one character comparison is one function call; it is not wrong, it counts a quantity that does not dominate wall time on this hardware. So this landed on correctness and less redundant work, and the changelist records that rather than claiming a win we did not observe.

Thank you -- and thank you for the whole batch. Ten PRs, nine landed as written or nearly so, and the diagnoses were consistently better than the patches needed them to be.

showell:
Thank you for catching this, and thank you for fixing the four call sites rather than bouncing the patch. **You are right and we sent a defect.** Two things below: what we should have done, which is embarrassing and specific, and a correction we owe you on the title.

## How we should have caught it, and it is not subtle

We have a cross-reference tool. Pointed at your tree, unmodified, it answers in about a second:

```
$ xref who apply-cite-overrides codex/
apply-cite-overrides: defined in Chapter Scoper (Semantics/ChapterScoper.codex)
  read by 3 chapters:
    Lowering               IR/Lowering.codex
    Name Resolution        Semantics/NameResolver.codex
    Type Checker           Types/TypeChecker.codex
```

That is your list. We wrote the tool, we did not run it on the one change in the batch where a signature's *meaning* moved, and the reason we did not is worth stating because it is the actual lesson: **we ran it out of the habit of asking "who calls this?" when we are reading, and not out of a rule that says "before changing a signature, enumerate the callers."** The type did not change — `List Text` before and `List Text` after — so nothing failed, nothing warned, and the question never got asked.

The rule we are adopting, and we would rather write it down publicly than promise it privately:

> **When a parameter's MEANING changes and its TYPE does not, the compiler is not a check and no arm is a check. Census the callers and read every one.**

## Why our arms said the patch was fine, which we think is worth your knowing

The PR body claims the IR is byte-identical over 6,890 definitions, checked on an arm that shares no code with the one that produced it. That claim was true and it was worthless here, and we would like to confirm we understand why.

`apply-cite-selected` acts only inside `if is-colliding colliding n`. A selected name that does not collide across chapters in the unit takes the else branch and the loop is a no-op — so a wrong `cite-slug` produces no wrong rename at all unless the unit contains a colliding selected name. If the compiler's own source has none, then both the correct hoist and our mis-indexed one emit exactly the same bytes, and every subject we own is blind to the difference.

If that is right, then:

1. **Is there a subject that DOES exercise it?** If one exists in your tree we would like to know its name, so we can make it a required arm for anything touching cite resolution. If none exists, that is worth knowing too — it would mean the path is currently untested on both sides, which is a finding rather than a complaint.
2. **Is a colliding selected name something the corpus is expected to contain**, or is it rare enough that this code path is mostly latent? Your answer changes how alarmed we should be that we broke it and nothing noticed.

## The correction we owe you on speed

> Emitting the compiler's IR takes 148.0s under the head compiler and 147.9s under this one -- 0.2s across 148s.

Accepted, and thank you for measuring it and saying so rather than letting the title stand.

The body was careful — "counted rather than timed, so the number is about the program and not about the machine" — and **the title was not.** "a third of a self-compile" reads as a wall-clock claim to anyone who does not get to the third paragraph, and a title is what most readers see. That is on us, and the fix is ours to make: **when the measurement is a proxy, the title carries the proxy's name.** This one should have read "a third of the interpreter steps of a self-compile", which is duller and true.

The earlier 12% from two 6-second self-compiles you correctly call noise. We should not have reported an n=2 difference on a 6-second workload at all; our own rule is that n=1 is a phenomenon and not a finding, and n=2 is barely better.

What we would still say for the record, hedged: the *second* measurement in this thread is not a step count. Peak linear memory on the self-hosting wasm module went 2180.2 MB to 1655.1 MB, which that repository gates on explicitly because wasm32's 4 GiB is an address space rather than a policy. On a bump allocator that never frees, avoided work and avoided memory are the same thing — which is why we believe the change is real even though it does not move your wall clock. We did not predict it and we do not claim it generalises to your hardware.

## And thank you for the batch note

> Ten PRs, nine landed as written or nearly so, and the diagnoses were consistently better than the patches needed them to be.

That is generous, and the one that did not land as written is the one we most needed to hear about. We look forward to testing all of it when Update 56 reaches us — none of this is re-verifiable from our side while the work is in Perforce, so this reply is a reading of your notes and not of your tree.


---

damiant3:
Shipped in Update 56: commit 6cd2ca1bd9dddca5e6358021bbc0fb5d6d5e874e on GitHub master and GitLab main (2026-09-08), carrying the landing recorded above (Perforce main 22500). Credit is in the release note, docs/PM/Active/GitHubUpdates/GitHubUpdate56.md, paragraph "Outside contributions". Thank you.

PR #133, opened 2026-09-06
COMPILER-54: fifteen foreword chapters are CRLF and cannot be read as written

showell:
Fifteen `.codex` files under `codex/foreword/{ai,engine,math,signal,sim}` have CRLF line endings. Every other `.codex` file in the checkout is LF — 0 of the remaining 3,700-odd carries a carriage return — so these are the exception, and have been since at least `53b3b213`.

**A reader that keeps the bytes cannot compile them, two different ways:**

```
codex/foreword/math/Complex.codex     16 x CDX1000
codex/foreword/signal/FFT.codex       CDX3007
```

`CDX1000` because a carriage return inside a **type** is refused. Measured one probe per position against `native/codexir`: an unmapped byte is skipped in an expression and on the lines around a definition, and halts the compile inside a type signature. Carriage return, tab, vertical tab, form feed and SOH all behave identically there — so this is the general rule, not a rule about CRLF, and it is reported here as context rather than as something to change.

`CDX3007` for a different reason, and it is the one that hides: a `Chapter:` header ending in CRLF carries the return into the chapter **name**. The chapter is present in the unit and `cites Signal chapter FFT` still cannot match it, because what is present is `FFT\r`. The unit then reports a missing chapter, which is not what is wrong with it.

**Why this has never bitten.** Every ordinary front door normalises on read — PowerShell's `Get-Content` does, and so does Python's `read_text`. The defect is invisible from the outside and appears the moment something reads the files faithfully. We hit it building a byte-preserving bundler.

**The change is whitespace only, and measured to be a no-op.** CRLF becomes LF, and a stray `\r\r\n` becomes one LF. All 35 corpus programs that cite one of these fifteen chapters were compiled through `native/codexir` before and after the change: **35 identical IR, 0 different.** No program changes meaning; what changes is that the files can be read by something that does not correct them first.

Not seed-affecting. Backlog row COMPILER-54 included.

Prepared by Claude (Anthropic), working with Steve Howell.

### DIFF
--- a/codex/compiler/compiler-backlog.md
+++ b/codex/compiler/compiler-backlog.md
@@ -210,3 +210,43 @@ the live `utf8-to-cce` path, so peeling them is a separate change.
 as the symbol being absent from compiler source. A recursive
 `Get-ChildItem | Select-String` over the 3,948 live `.codex` files is what
 answered it.
+
+## COMPILER-54 -- CONTRIBUTED by Steve Howell (PR pending, 2026-09-06): fifteen foreword chapters are CRLF, and the compiler cannot read them as they sit on disk
+
+`codex/foreword/{ai,engine,math,signal,sim}` hold fifteen `.codex` files with
+CRLF line endings. Every other `.codex` file in the checkout is LF -- 0 of the
+remaining 3,700-odd carries a carriage return -- so these are the exception, and
+they have been since at least `53b3b213`.
+
+**A reader that keeps the bytes cannot compile them, two different ways.**
+
+    codex/foreword/math/Complex.codex     16 x CDX1000
+    codex/foreword/signal/FFT.codex       CDX3007
+
+CDX1000 because a carriage return inside a TYPE is refused. Measured one probe
+per position against `native/codexir`: an unmapped byte is skipped in an
+expression and on the lines around a definition, and halts the compile inside a
+type signature. Carriage return, tab, vertical tab, form feed and SOH all behave
+the same way there, so this is the general rule and not a rule about CRLF.
+
+CDX3007 for a different reason and it is the one that hides: a `Chapter:` header
+ending in CRLF carries the return into the chapter NAME. The chapter is present
+in the unit and `cites Signal chapter FFT` still cannot match it, because what is
+present is `FFT\r`. The unit then reports a missing chapter, which is not what is
+wrong with it.
+
+**Why this has never bitten.** Every ordinary front door normalises on read --
+PowerShell's `Get-Content` does, and so does Python's `read_text`, which is how
+our own harnesses reach the compiler. The defect is invisible from the outside
+and appears the moment anything reads the file faithfully.
+
+**The change is whitespace only, and measured to be a no-op.** CRLF becomes LF
+and a stray doubled `\r\r\n` becomes one LF. All 35 corpus programs that cite one
+of these fifteen chapters were compiled through `native/codexir` before and
+after: **35 identical IR, 0 different.** So no program changes meaning; what
+changes is that the files can be read by something that does not correct them
+first.
+
+Not seed-affecting.
+
+Reported and prepared by Claude (Anthropic), working with Steve Howell.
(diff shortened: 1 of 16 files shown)


damiant3:
Landed as **COMPILER-54**, with credit -- but fixed differently from the patch, and the difference is worth writing down because it is about our depot rather than about your diagnosis.

Perforce depot: blu 22463 and 22465, on main as **22466**.

**Your diagnosis is exactly right and the CDX3007 half is the valuable part** -- a CR at the end of a `Chapter:` header puts the CR in the chapter NAME, so the chapter is present and `cites Foreword chapter FFT` still cannot match it, and the unit then reports a missing chapter, which is not what is wrong with it. That is the kind of defect that costs someone a day.

**What differs is the size of the fix here.** Perforce already normalises the line terminator on this tree, so the fifteen files are not CRLF in our depot: what survives is **one stray CR per file** (two in `signal/Noise.codex`), 20 bytes in total, each sitting at the end of a `Chapter:` header. Removing those is one changed line per file. Converting CRLF to LF would instead rewrite every line of fifteen files -- 143 lines in `FFT.codex` alone -- against a tree whose working files are CRLF by client setting.

**The set is 19, not 15.** You looked at `codex/foreword`. A census over all 3,884 `.codex` in the tree found four more: `apps/globe/Relativity.codex`, `apps/mathbook/Statistics.codex`, `apps/spark/SparkIntro.codex`, `apps/spark/SpectrumAnalyzer.codex`.

**And no Perforce setting causes or prevents it**, which was worth establishing because it is the obvious guess: the fifteen and their controls are all filetype `unicode` on a client whose `LineEnd` is `local`. Translation rewrites the terminator; a CR that is not one is content and is preserved exactly. So the only thing that can keep this out of the tree is a reader, and the CL adds one -- `build/check-source-cr.ps1`, calibrated in both directions on the real population (19 of 3,884 before, 0 after) and reporting the line rather than the byte offset.

Thank you. The mechanism was right, the CDX1000/CDX3007 split was right, and the only thing that did not transfer was an assumption about how our depot stores these files, which you had no way to see from outside.

showell:
Thank you — and thank you for writing down *why* the fix differs. That is the most useful comment in the batch for us, because none of it was visible from outside and all of it will change what we send next time.

Three things we have taken from it:

**1. Your depot stores these files differently than our clone shows them.** Perforce normalises the terminator, filetype `unicode`, client `LineEnd` local — so what we saw as fifteen CRLF files is, on your side, twenty stray CR bytes. A patch that rewrites every line of fifteen files against a tree whose working files are CRLF by client setting is a bad trade for a 20-byte defect, and we could not have known that. **Is there anything else about depot storage that a patch from outside should account for?** We would rather learn the set than discover them one PR at a time — filetypes, keyword expansion, anything where the bytes we see are not the bytes you have.

**2. The census should have been tree-wide, and that one is entirely ours.** We looked at `codex/foreword` because that is where the failure surfaced, and the real population is 19 across 3,884 files. There is no excuse: the grep costs nothing and we had already decided the defect was about a byte rather than about a directory. We will census the whole tree for any defect whose signature is textual, and we will report the denominator (`19 of 3,884`) rather than just the numerator.

**3. "No setting causes or prevents it, so the only thing that can keep it out is a reader"** is a good rule and we had not reached it. Calibrating `build/check-source-cr.ps1` in both directions on the real population — 19 before, 0 after — is the discipline we try to hold for our own gates and did not apply to this proposal. We proposed a fix and not a guard.

We look forward to testing it when Update 56 reaches us. And thank you for keeping the CDX3007 half attached to the diagnosis — a CR landing inside the chapter NAME, so the chapter is present and the cite still cannot match it, is the part that costs someone a day, and it reads clearly in your changelist.


---

damiant3:
Shipped in Update 56: commit 6cd2ca1bd9dddca5e6358021bbc0fb5d6d5e874e on GitHub master and GitLab main (2026-09-08), carrying the landing recorded above (Perforce main 22466). Credit is in the release note, docs/PM/Active/GitHubUpdates/GitHubUpdate56.md, paragraph "Outside contributions". Thank you.

PR #134, opened 2026-09-08
compiler: a definition's emitted IR type is decided by where an unrelated definition's NAME sorts

showell:
`all-bindings` holds **two entries per definition**, and a lower-bound search
over it returns whichever one a non-stable sort happened to leave first.

The double entry is deliberate and you document it. `CheckHarness.codex`:

> `type-bindings` IS `all-bindings`, WHICH IS BIGGER THAN IT LOOKS AND
> DOUBLE-COUNTS. It is `check-result.types & resolved-env`, so a chapter's own
> defs appear TWICE -- once inferred, once from the environment.

with a worked count for bare fib: 257 builtins + 3 defs from the environment +
3 defs again from `check-result.types` = 263.

**What does not look deliberate is that the two entries DIFFER for a
polymorphic definition, and the lookup picks between them by accident.**

## The chain, all of it read rather than inferred

1. `compile-checked` builds
   `pre-sorted = sort-bindings (check-result.types & check-result.env.bindings)`.
2. The two entries for one definition are not the same type. The environment's
   is `parameterize-type`'s output and carries a `ForAllTy`;
   `check-result.types`' is `check-def-normal`'s `inferred-type`, which is
   `instantiate-collect (env-lookup env rn)` and has the quantifier peeled.
3. `sort-bindings` is `sort-by xs compare-binding-names`, and
   `compare-binding-names` compares **the name and nothing else**.
4. `sort-by` is `qsort-by` -- median-of-three, Lomuto partition -- and it is
   **not stable**. For a run of equal keys the survivor at the front is decided
   by the final `sort-swap xs i hi`.
5. `lookup-type-bsearch` uses `bsearch-text-pos`, which is a **lower bound**, so
   it returns the LEFTMOST of the equal run.
6. `lower-def` takes `type-val = deep-resolve (lookup-type-bsearch types rn)`.

So the emitted type of a definition is whichever of its two bindings the
quicksort left leftmost.

## What it does at head

Of the **104** definitions in the compiler's own IR whose type mentions a type
variable, **103 carry the inferred type and exactly one does not**:

```
(def "deck-record" "Parsmi--Phase Allocator" (params (param "x" (tvar 86)))
     (forall 86 (fn (tvar 86) (tvar 86))) (name "x" (tvar 86)) 0 0)
```

`map-list`, `qsort-by`, `codex-type-fold-children` and the other 102 are
spelled without it. Nothing in the source distinguishes `deck-record` from
them.

## Demonstrated, not argued

Five **unreachable** definitions added to Chapter Phase Allocator, chosen only
so their names sort next to `deck-record`:

```
  deck-recora : Integer -> Integer
  deck-recora (n) = n + 1
  ... deck-recorb, deck-recorz, deck-records, deck-recordz
```

The quantifier disappears.

```
emitted IR      8,141,768  ->  8,141,756 bytes
the difference  12 bytes, exactly `(forall 86 ` and its `)`
definitions     2,845 of 2,846 byte-identical
the five added names       DCE'd; they appear nowhere in the output
type-variable ids          unchanged, 86 on both sides
```

**Perturbations that sort far from it change nothing.** One unrelated
definition: byte-identical. Fifty: byte-identical. A second polymorphic
identity function `zz-ident : a -> a`: byte-identical, and it does not acquire
a quantifier of its own. That is why this survives -- it takes a perturbation
aimed at the same region of the sorted list to move it, and nobody aims one.

## Why we are not proposing a fix

Three repairs are available and each costs something only this tree can price.
We can measure none of them from outside, and this is core-compiler code we
have no way to test here -- no bare metal, no BVT, no fixed point.

1. **Deduplicate `all-bindings`.** The obvious repair, and it changes
   `type-bindings` -- a number your own harness publishes, documents and grades,
   with a worked example. Every harness dump moves.
2. **Make `sort-bindings` stable for equal keys.** The smallest semantic change,
   and it keeps the count: the inferred entry comes first in the concatenation,
   so a stable sort makes it win deterministically. But `sort-by` is a Foreword
   primitive and everything uses it.
3. **Have `lower-def` read `check-result.types` rather than the union.** The
   narrowest, and it leaves every other consumer of `all-bindings` reading the
   same ambiguous table.

## How this was found, and one corroboration

By building an independent front end and diffing 8.1 MB of the compiler's own
IR definition by definition. 2,845 of 2,846 matched; this was the last one.

That front end answers the **inferred** type. So it agrees with this compiler
on **2,846 of 2,846** definitions once the fragile case is perturbed away --
which is the strongest evidence available from outside that the inferred entry
is the answer you intend, and the environment's entry is the accident.

We had four wrong theories before this one, including two we were confident in,
so please read the demonstration rather than the reasoning: the 12 bytes and
the five inert definitions are the part that does not depend on us being right
about your compiler.

https://claude.ai/code/session_01VQfsk5tuZj9uEYU3hLUqzp


damiant3:
Confirmed at head, and your instance has already moved, which is a better demonstration than the one you shipped.

Ingested as **COMPILER-66** in `codex/compiler/compiler-backlog.md`, credited to you and to this PR.

## What we verified

Every link, by reading the source rather than inferring it:

- `opening.codex:652` -- `pre-sorted = deck-record (sort-bindings (check-result.types & check-result.env.bindings))`. The concatenation order is types first, environment second, exactly as you say.
- `IR/LoweringTypes.codex:51` -- `compare-binding-names (a) (b) = text-compare (a.name) (b.name)`. The name and nothing else.
- `foreword/core/Sort.codex:13-28` -- `sort-by` is `qsort-by`, median-of-three, in place through `sort-swap`. Not stable, and the chapter never claims to be.
- `Core/Collections.codex:4-11` -- `bsearch-text-pos` is a lower bound, and the section header says "lower bound" in those words.
- `IR/LoweringTypes.codex:34-41` -- `lookup-type-bsearch` returns `list-at bindings pos`, the leftmost of the equal run.
- `Ir/Lowering.codex:1646` -- `raw-type = lookup-type-bsearch types rn`.

## What we measured

We emitted the compiler's own IR at head and looked for the symptom.

`deck-record` is now clean:

```
(def "deck-record" "Core--Phase Allocator" (params (param "x" (tvar 93))) (fn (tvar 93) (tvar 93)) (name "x" (tvar 93)) 0 0)
```

No quantifier, and the type-variable id is 93 rather than your 86. Nobody fixed anything and nobody aimed a perturbation at it. Ordinary landings shifted the region.

Three other definitions now carry the environment's entry instead, out of 90 whose type mentions a type variable:

```
hamt-node-set    [Foreword--Hamt]        forall 35
sort-by          [Foreword--Sort]        forall 60
copy-list-loop   [Types--Unification]    forall 99
```

`sort-by` is the one worth pausing on. **The unstable sort at the heart of this is itself one of the three definitions currently spelled with the quantifier, while its own helper `qsort-by`, four lines below it in the same chapter, is spelled without one.** Nothing in the source distinguishes them.

So the count moved from 1-of-104 to 3-of-90 with no intent behind either number. That is your point about survival, demonstrated a second time by a tree you cannot see.

## On the three repairs

Your reading of the trade-offs matches ours, including which one is narrowest. Two notes from inside:

Option 2 is the one we are most interested in, and your observation that the concatenation already places the inferred entry first is what makes it work: stability alone decides it correctly, with no change to `type-bindings` and no change to any published count. The cost you could not price is that `sort-by` is a Foreword primitive with callers across the tree and the plugs, so making it stable is a change to a shared contract rather than to a compiler internal, and it wants its own measurement of what stability costs on the large sorts.

Option 1 would move `type-bindings`, which `CheckHarness.codex` publishes, documents and grades with the worked count of 263 for bare fib that you quote. That is a fixable cost rather than a blocking one, but it means the harness change and the compiler change land together.

We have not chosen yet. The row records all three with your pricing.

## On method

Declining to propose a fix you could not test was correct, and the demonstration carried the report. The five inert definitions and the 12 bytes are what made this reproducible for us before we had read a line of your reasoning, and telling us you had four wrong theories first, two of them confident, is why we read the demonstration first.

Closing as ingested, with thanks. The defect is ours to fix and the row is open.

damiant3:
Ingested as COMPILER-66. Verified at head and the row is open; see the review comment above for what moved.

showell:
Thank you — and the second demonstration is better than ours in the way that
matters: nobody aimed it. `deck-record` clean at tvar 93, three others carrying
the quantifier instead, 3-of-90 where we saw 1-of-104, and no intent behind
either number.

**`sort-by` spelled with a quantifier while `qsort-by` four lines below it is
not** is the example we would have wanted and could not have found. The unstable
sort wearing its own defect, in the chapter that defines it.

## Two numbers for pricing option 2, since you named the gap

You said stability wants its own measurement of what it costs on the large
sorts. There are eight `sort-by` call sites in the compiler and two dominate a
self-compile. Both counts are exact — they are counters we grade against
`codexcheck` and they match:

```
sort-expr-types    145,772 entries      the expr-type table
sort-bindings       13,646 entries      all-bindings
```

The other six are small and bounded — x86 emitters, x86 arities, zig builtins,
rename dedupe. So the question is really about one sort of ~146k and one of
~14k, each once per compile, and not about a hot path.

## And a cost that may not be on the list

**A stable sort usually wants O(n) scratch, and this allocator never frees.** On
a self-compile that is 145,772 entries of scratch appearing inside the check
phase, which is exactly the shape that moves peak memory rather than time.

That is measurable, and the instrument is already banked: `codex-wasm-transpiler`
builds a wasm module that compiles the source that produced it, and **it gates
on peak linear memory explicitly** — it refuses a build more than 2% over its
banked value, because wasm32's 4 GiB is an address space rather than a policy.
It is the arm that showed PR 132 moving 2180.2 MB -> 1655.1 MB when nothing else
we had could see that change at all.

**If you produce a candidate with a stable `sort-by`, we will run it and report
the peak.** That is a number you cannot get from a wall clock and we can get
cheaply. Say the word and we will; no need to hand us anything but a branch.

## On the harness cost of option 1

Worth adding, since we quoted the number at you: `CheckHarness.codex` publishes
`type-bindings` as `all-bindings` and explains the double count with the worked
263 for bare fib. If option 1 lands, that prose is part of the change and not
just the figure — the explanation is what makes the number readable, and a
deduplicated list would make the paragraph wrong rather than merely stale.

Thank you for reading the demonstration first. We will keep sending them that
way.

---

damiant3:
Answered late, sorry: your note landed while the emitted half was mid-chain.

**The emitted half landed on main 23732 by a fourth repair, narrower than the three you priced.** `lower-def` already computed `stripped = strip-forall-ty full-type` and used it for params, return, tvar ids and ctx; the one field still carrying the quantifier was `type-val`, the field that gets emitted. It now reads `type-val = stripped`, so both entries strip to the same shape and the sort order stops mattering for emission. Census at head: definitions emitting a forall in the type position 3 before, 0 after; definitions whose type mentions a tvar 90 on both sides, so the binder went and no type information did. Fixed point held, BVT 143 of 143, seed `12B76F40E611053F` at the same size. Consumer evidence for stripping over stabilising: arm64, riscv and wasm have zero `ForAllTy` arms, and the zig emitter's own prose says a quantified type had no rendering.

**What stays open on COMPILER-66 is the larger half:** `all-bindings` is still ambiguous for the ~20 readers that reach it through `lookup-type-split`, some of which strip for themselves and some of which do not. If that half is repaired by a stable `sort-by`, your peak-linear-memory arm is exactly the instrument it wants, and we will send a branch and take you up on the offer then. Your point about the `CheckHarness` prose (the worked 263 becomes wrong, not stale, under a deduplicated list) is on the row.

damiant3:
Shipped in Update 56: commit 6cd2ca1bd9dddca5e6358021bbc0fb5d6d5e874e on GitHub master and GitLab main (2026-09-08), carrying the landing recorded above (Perforce main 23732). Credit is in the release note, docs/PM/Active/GitHubUpdates/GitHubUpdate56.md, paragraph "Outside contributions". Thank you.

Issue #121, opened 2026-09-04
address-of refuses on a by-value union: the lowering memo-copy cannot be transpiled, and 6 of 14 ladder rungs fail on it


showell:
## What happens

Transpiling the compiler's own lowering to zig refuses:

    lower.zig:12018:17: error: zig plug: no address-of for lower.IRPat
    lower.zig:12018:17: error: zig plug: no address-of for lower.IRActStmt

The call sites are the lowering memo-copy family, e.g. `lcopy-pat`:

    return (if ((cx_address_of(p_) < mc.mc_floor)) p_ else switch (p_) { ... });

`address-of` there is asking whether the object is already below the memo floor
and can be shared instead of re-copied.

## Why bare metal does not see it

`address-of` on bare metal is `emit-identity-builtin`: it returns the VALUE, and
since records, lists and texts are pointers in one flat region, the value IS the
address. It is total by construction and no type can fail it.

A source plug has to reconstruct that meaning per type. The zig plug does:
integers return themselves, pointers return a heap-relative offset, and a
payload-free union returns its tag. A tagged union WITH payloads has no address
to give, so it refuses rather than answer something.

## The asymmetry is in the representation, and it is visible in the output

`IRExpr` is emitted BOXED and works -- `switch (e_.*)`, so `address-of` reaches
the `.pointer` arm and returns an offset. `IRPat` and `IRActStmt` are emitted BY
VALUE -- `switch (p_)` -- and have no address at all. Same generic code, same
call, three types, two representations.

## Why we think this is the same shape as COMPILER-43

The comparison is bare-metal memory management: an address against a deck floor,
deciding whether a copy can be shared. A hosted target has no such memory --
its objects live wherever the host allocator put them -- so the question is not
merely hard to answer there, it is not meaningful. That is the argument this
project already accepted for `check-batch-close` (COMPILER-43 / our issue 117),
where the whole operation is guarded by `hosted-kind`.

If that reading is right, the fix is a guard rather than a new emitter arm, and
`address-of` on a by-value union stays honestly refused.

## What it blocks

On the plug-oracle ladder at Update 55 plus the zig-plug memory builtins:
6 of 14 rungs green. `lex`, `parse`, `desugar`, `scope`, `check` and
`lir_to_x86` are byte-identical to bare-metal truth. `lower`, `ir_to_codex`,
`ir_to_codex_roundtrip`, `ir_to_wire`, `ir_to_x86` and `passes_to_x86` all fail,
[...]

PR #135, opened 2026-09-09
zig plug: box every payload-carrying variant, so address-of is total

showell:
direction. This is the fix for the defect we filed as #121 and then withdrew,
saying it is ours to fix in the zig plug rather than yours to guard in the
compiler.*

## What this is

Bare metal boxes every variant value: they are heap words, which is why
`address-of` there is `emit-identity-builtin` and no type can fail it -- the
value **is** the address. This plug boxed only what zig's finite-size rule
forced, so `address-of`'s meaning rested on a decision made for SIZING:
`IRExpr` held `IRExpr` directly and was boxed, `IRPat` reached itself only
through `List IRPat` and was not, and `cx_address_of` refused on the second.

This boxes every payload-carrying variant, so the plug agrees with bare metal's
own model and `address-of` is total the way it is there. A generic variant
stays unboxed -- it is emitted as a `fn ... type` with no single declaration to
point at.

## Why it matters

The gap was always present; nothing reached it until Update 55's compiler
memory campaign added the lowering memo-copy (`lcopy-pat`, `lcopy-stmts`) and
asked `address-of p < mc.mc-floor` of an IR pattern for the first time.

- **Update 55:** six of fourteen ladder rungs fail, every one on this refusal.
- **Update 56:** it stops the zig build of the compiler itself -- the plug
  cannot emit a compiler that copies its own IR patterns.

Our Update-58 candidate carries this fix and self-hosts through the zig plug;
it also passes our corpus, self-host, safari and WGSL gates unchanged, so the
change is contained to the boxing decision and does not move emitted bytes
elsewhere.

## The change

Three commits, all in `codex/plugs/zig/ZigEmitter.codex` (net +27 / -14):

1. `zig-typedef-recursive` becomes `zig-typedef-boxed` -- box every
   payload-carrying variant, not only the self-recursive ones; a generic
   variant stays unboxed.
2. a one-line follow-up: the real-to-text prelude was shadowing the prelude's
   own boxing helper.
3. a comment stating the invariant rather than the discovery, so it survives
   ingest.

Expressed against the single-file `ZigEmitter.codex`. On our own branch this
lives across three emitter pages; that four-page split is not upstream and is
deliberately not part of this candidate.

### DIFF
--- a/codex/plugs/zig/ZigEmitter.codex
+++ b/codex/plugs/zig/ZigEmitter.codex
@@ -1314,7 +1314,7 @@ Section: Name Mapping
    in if ci >= 0 then
     let c = list-at (ctx.ctors) ci
     in if c.enum-like then zig-sanitize (c.type-name) & "." & zig-sanitize n
-       else let boxed = zig-typedef-recursive ctx (c.type-name)
+       else let boxed = zig-typedef-boxed ctx (c.type-name)
        in let base = if boxed then zig-sanitize (c.type-name) & "S" else zig-sanitize (c.type-name) & zig-ctor-type-args ty (ctx.scope-tvars)
        in if boxed then "cx_new(" & base & "{ ." & zig-sanitize n & " = {} })"
           else base & "{ ." & zig-sanitize n & " = {} }"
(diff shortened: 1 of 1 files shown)


damiant3:
Received, and queued directly behind PR 137. Note for sequencing rather than a problem with the change: our lane is about to edit ZigEmitter.codex for a different reason, replacing zig-p-cx-text-to-double-bits, which the 125 fix left disagreeing with the seed. We are doing that port and your two zig PRs in one lane so the file does not get merged three ways. You also noted this resolves the runtime face of issue 126 (COMPILER-56 on our side); we will confirm that when it lands rather than take it on faith.

damiant3:
Review resumed in the Zig lane. The decimal-parser port mentioned above has now landed in Perforce, and we will preserve that code when applying the boxing change. We will also verify the six hosted comparison types from issue 126 before treating that issue as resolved. Thank you, Steve; the PR remains open until the accepted work is public.

damiant3:
The focused PR 135 controls now pass: eight address cases, the plain-sum regression and all 40 rounding rows. Measured allocation for the tested payload variants is 1,600 bytes for 100 objects and 3,207 bytes for 200 including alignment, so the added boxing cost is explicit.

Two broader Zig checks remain red: eq-generic-fields emits undeclared a_ type arguments, and typeclass-smoke carries undeclared T578. Both failures also occur with the old plug. That comparison identifies pre-existing failures, but does not provide passing coverage for those checks. We are holding the copy-up and checking both against the PR 139/140 Lowering candidate before deciding the remaining scope.

The full hosted comparison driver for issue 126 is separately blocked by port-in-byte, as reported there. No full-hosted or issue-126 success is claimed.

damiant3:
The diagnosis now localizes both primary blockers upstream of Zig. Generic equality calls carry concrete arguments, but the synthesized helper's parameter and match annotations retain formal a/b types. The typeclass case likewise reaches the wire with an unresolved dictionary type where the call expects a concrete instance. A separate blanket parameter-discard behavior in Zig adds a secondary diagnostic when a refusal is nested in a body.

We are repairing the equality annotation path and investigating structural instance-type propagation in separate compiler units. The instance declaration hypothesis still needs its controlled test; no global defaulting of free variables is planned. The same frozen old and PR135 plugs will grade the repaired IR and full expected outputs. PR135 remains held until those checks pass, while PR138/141 preparation proceeds in preserved dependent shelves.

damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). Landed in Perforce main 25656 together with
PR 138 and the Zig half of PR 141, with credit to you: every payload-carrying
variant is boxed, so address-of is total, and the `cx_frac` reservation is
included. Six exact Zig controls pass on the seed. Two Zig reds you reported
are upstream of the plug and stay open on our side: `eq-generic-fields`
(generic equality helpers keep formal parameter types) and `typeclass-smoke`
(an unresolved dictionary type), tracked as plugs-backlog 2.54 and the
MethodSpecialization design. Nothing further is owed from you. Thank you.

Issue #125, opened 2026-09-04
Real literals are not correctly rounded: 10 of 120 ordinary doubles land one ULP away, in the front end


showell:
This is Claude, working with Steve. Found when two of our unit-test chapters passed on our own Codex interpreter and failed on the zig plug.

## It is the literals, not the arithmetic

We probed with `real-to-bits`, since `show` on a `Real` is refused and it is the only instrument that can compare a Real across arms. **Every computed value in both tests is bit-identical on the two arms.** What differs is the literals they are compared against.

## Measured

120 random doubles in (-1000, 1000), each written as its shortest round-tripping decimal, converted and read back as bits:

| front end | wrong | worst gap |
|---|---|---|
| an independent Rust front end | **0 of 120** | — |
| Codex | **10 of 120** | 1 ULP |

Three concrete cases — correctly-rounded value first:

```
0.012500000000000011   4578359381184846240    Codex gives ...241
0.40800000000000003    4601021494509774570    Codex gives ...569
0.42350000000000004    4601300717686671541    Codex gives ...540
```

## It is the front end, so every backend inherits it

`codex/plugs/zig/ZigEmitter.codex:1616` renders a Real literal as

```
"@as(f64, @bitCast(@as(i64, " & integer-to-text n & ")))"
```

— it prints a bit pattern it is handed and does no conversion. And `codex/compiler/IR/IRChapter.codex:21` declares `IrNumLit (Integer) (SourceSpan)`, typed `real-f64` at line 173. **The IR carries the bits**, so the decimal is converted at or before IR construction and bare metal, wasm and the plugs are all handed the same value. Zig's own float parser never sees the decimal.

## What we did not establish

We did not find the conversion routine — we searched `codex/compiler/Syntax` and the obvious digit-accumulation shapes without locating it, so we cannot say whether this is `acc * 10 + d` in binary, a power-of-ten scaling, or something else. We also did not run bare metal; the claim that it shares the defect is read from the IR declaration above, not measured.

## Why it is worth more than one ULP sounds

It is silent, it hits ~8% of ordinary literals, and it puts a floor under any cross-arm test that grades a computed Real against a written-down Real at tolerance 0.0 — such a test can fail on an arm for reasons that have nothing to do with the arm. The usual fix is a correctly-rounded decimal→binary conversion (Clinger / Eisel-Lemire style), or at minimum arbitrary-precision accumulation of the significand before a single final rounding.

[...]

PR #137, opened 2026-09-09
test/ops/real-literal-rounding: boundary cases for COMPILER-57 (red at head until the fix)

showell:
direction. This is the boundary-cases test you okayed on issue #125.*

## What this is

It extends `codex/test/ops/real-literal-rounding` with five boundary literals
and their correctly rounded `.expected`. The existing twelve lines are untouched.

**Three of the five are RED AT HEAD on purpose**, and that is the point. The
divisor fix this chapter records is correct, but the significand conversion is
not: the digits accumulate into an i64 and `cvtsi2sd` makes that an f64, which
is exact only below 2^53, so a literal needing 16 or more significant digits
rounds twice and lands one ULP off. The `.expected` values are the *correctly
rounded* bits, so `11.700000000000001` and `97.59752277630605` (one ULP low) and
`993.1027217047139` (one ULP high) mismatch on today's compiler. The two
controls stay green: `0.123456789012345` is 15 digits, and `5.8500000000000005`
is 16+ but rounds the right way — it is here to show the defect is
data-dependent, since its sibling `11.7` does not.

Per your note, this lands with the parser fix (COMPILER-57), not before — a
chapter red at head blocks the gate, so it waits on your side until the parsers
round once.

## Provenance of the expected values

The `.expected` bits are the correctly rounded conversion, taken from an
independent front end and cross-checked against Python's `float`, never captured
from the current compiler — capturing would write the bug in as the oracle.
Confirmed both directions: run through a correctly rounded front end all
seventeen lines match; run through today's compiler exactly the three 16+-digit
lines differ, two low and one high.


damiant3:
Taken first, ahead of the other three. The bare-metal fix for 125 landed (seed D50584DFBA0DCB80), so this should be green at head now rather than red by design. Your rows pin the correctly-rounded bits better than the twelve small values we had, so where yours and ours overlap we are taking yours. The compiler lane is ingesting it now.

damiant3:
Thank you, Steve. All five boundary literals and their expected bits are now present in Perforce. Three overlapped existing tests; the two missing rows, 0.123456789012345 and 993.1027217047139, landed in main change 25528. The complete merged rounding output passed against the corrected seed. We retained the broader internal regression cases and omitted the historical source prose about the old failing implementation.

Update 58 contains the three overlapping rows but not the two added rows. This PR therefore remains open until publication; the release update will link the public commit carrying all five.

damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). All five boundary literals and their
expected bits are in the tree; the two rows Update 58 lacked
(`0.123456789012345`, `993.1027217047139`) landed in main 25528. The merged
rounding output passes on the corrected seed. Thank you.

PR #139, opened 2026-09-10
inline-single-caller: the declared return type takes the call site's type

showell:

## The defect

`inline-single-caller` copies a helper's body into its one caller and recovers the helper's type variables by matching its **parameter** types against the argument types (`once-retype`). A variable that lives only in the **return** type has nothing to match, so the copy carries the helper's own quantified variable into the caller's wire:

```
  make-empty : Integer -> List a
  make-empty (n) = []

  opening : Integer = list-length (list-push (make-empty 0) "a")
```

At `49fa9f27` the inlined literal in `opening` is `(list-expr (elems) (tvar 2))` — `make-empty`'s `a`, bound by nothing in `opening` — whether the site is `List Text` (above) or `List Integer` (`list-push (make-empty 0) 1`). The checker knew the difference (the mixed program is refused, CDX2001); the wire dropped it. This is COMPILER-30's half two in a new producer: not a `synthetic-span` node, but an optimizer pass discarding the type the site already carried.

The guard that skipped retyping when no *parameter* was polymorphic (`any-tyvar-param`) is the actual miss — it never looked at the return type.

## The fix

`InlineLeaf` carries `il-rty`, the declared type after the parameters (`peel-applied-return`, already in `CodexType Helpers`). `once-apply-site` pairs it with the site's own type — `ir-expr-type e`, the checker's resolved answer for exactly that node — as one more parameter/argument pair through the existing `once-retype`, and the guard reads the whole signature. The pair lists are built fresh: `list-push` writes in place and `il-ptys` is the candidate's template.

Not a sweep: only variables the candidate's signature names are touched. A site whose own type is still a variable passes that variable into the copy, which is the correct answer for a generic caller.

After the fix the two cells read `text` and `int-default`.

## Pinned by an `ir-fidelity` case

`build/ir-fidelity/cases/inlined-return-type` — `a`/`b` differ in the pushed element, `knows` mixes them and is refused with CDX2001, path `def:opening/body/find:list-expr/slot/2`. **It reads with `-Passes`**: the literal is in `opening` only after the pass runs, so under the default `-Passes none` it is UNSUPPORTED (the case's comment says so). Measured through a hosted `codexir` (this fix on our u58 stack, below): DROPPED before, CARRIED after.

I did not add a `codex/test` unit: bare metal erases the type and prints the right answer with or without the fix, so an `.expected` unit for this cannot go red.

## What the fix does not do — COMPILER-74, registered in this PR

When nothing constrains the site either (`list-length (make-empty 0)` alone), the copy now carries the **site's** variable instead of the helper's — the same `(tvar 6)` `list-length`'s parameter carries, so the def is internally consistent — but a typed plug still refuses (`zig plug: no element type for this empty list`), and bare metal prints `0`. That is a checker gap, not the inliner's: no per-definition finalization defaults a variable no constraint reaches. Our Rust front end has that phase (`default_ambiguous_vars`) and emits `int-default` there, which the zig plug builds. The row carries the specimen and the reading; the number is provisional to your register. `codex/test/roc-alias-empty` fails the zig plug today for the same reason.

## Measurement

Stack: `u58-candidate` = Update 57 `49fa9f27` + PR 135 + PR 138 + this commit (`527d438f`); the PR branch itself is off clean `49fa9f27`. `master` has not moved `IR/Lowering.codex` since Update 57.

- **Self-host, no guest** (`bootstrap_native.py`): converged in ONE round — the predecessor and the rebuilt `codexzig` emit byte-identical zig (2,964,156 bytes) for the compiler's own source. The compiler contains no single-caller helper of this shape.
- **Fixed point under QEMU** (`build.py --force`, the seed-compiled emitter against the native build): **HOLDS, byte-identical** (`build.py --force` on the same stack, 537 s, three guests; `arith` matches all 9 lines).
- **Corpus census, hosted `codexir` before vs after, all 1,269 programs of our u58 unit corpus:** **1,052 identical, 0 differ**, 217 refused by both (the same 217). Not one byte of IR moves anywhere in the corpus: no program in it inlines a single-caller helper whose type variable lives only in its return type. The change is confined to programs of the specimen's shape, and the census says how rare that shape is today.
- Zig plug arms unchanged: the 28 Cobblestone-curated programs 28/28 match; the 29 roc-lang ports 25/29, the same four failing before and after (`roc-alias-empty` is COMPILER-74; the three `roc-iter-*` are an unrelated unresolved variable).
- `codexcheck` counters on the compiler's own source: unchanged (`check-errors 0`, this pass runs after the checker and cannot move them).

Risks: this is `IR/Lowering.codex`, core compiler; measured on the hosted arms above and not on bare metal. The change is confined to `once-apply-site` and the candidate record; the leaf and cost-based inliners construct the record and read nothing new.


damiant3:
Thank you, Steve. We have read the return-type diagnosis and are taking PR 139 followed by PR 140 through one compiler lane, because both touch Lowering. We will reproduce the dropped-versus-carried IR result with the optimizer enabled, check the per-candidate allocation cost, and prove the resulting seed before landing in Perforce. The unconstrained empty-list case remains a separate limitation; we will map the provisional backlog number to our current register.

We will keep the PR open until the Git release carries the accepted work, then link the public commit and describe any changes made during ingestion.

damiant3:
Accepted and landed in Perforce main 25585, together with PR 140, with credit to Steve Howell. The IR-fidelity case discriminates: the baseline drops the return-only type variable; the candidate carries text and int-default at the two sites. Native runtime output alone does not distinguish those arms, so the IR check remains the acceptance evidence.

The landed signed seed matches the proven candidate byte for byte. Fixed-point, five focused runtime checks, BVT (79 compile, 64 runtime and batch), and signed self-verification passed. We preserved the newer list-equality changes in Lowering. The unconstrained empty-list case is recorded separately as COMPILER-81; its typed-Zig refusal is locally confirmed and is not claimed fixed by this PR.

Thank you, Steve. The Git mirror has not received main 25585 yet. This PR remains open until the release carries the work and we can link the public commit.

damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). Landed with PR 140 in main 25585, with a new
signed seed: an inlined single-caller definition's declared return type takes
the call site's substitution, and your IR-fidelity case distinguishes the old
unresolved variable. One adjacent gap stays open as COMPILER-81: an
unconstrained empty-list element type still reaches typed IR, and Zig refuses
it. We did not add a default type. Thank you.

PR #141, opened 2026-09-10
zig and wasm plugs: a nested sub-pattern is a test, not a wildcard

showell:

## The defect

Both hosted plugs read a constructor or literal pattern **inside** a constructor pattern as `_`, and the zig plug cannot match a Text literal at any level. Two programs (the two Roc ports that found this), both with IR that `codexir` at `49fa9f27` and the Rust reference compiler we maintain emit byte-identically, and that the interpreter answers correctly:

```
  Logic = | LTrue | LFalse | And (Logic) (Logic) | Or (Logic) (Logic) | Not (Logic)

  shape-check (v) =
   when v
    is And (Or (LTrue) (LFalse)) (Not (LFalse)) -> "ok"
    is otherwise -> "bad"

  shape-check (And (Or LTrue LTrue) (Not LFalse))   -- interpreter: bad    zig plug: ok    wasm plug: ok
```

```
  Res (a) (e) = | Ok (a) | Err (e)

  describe : Res (Res Integer Text) Integer -> Integer
  describe (v) =
   when v
    is Ok (Ok (n)) -> n
    is Ok (Err (ignored)) -> 0 - 1
    is Err (code) -> code * 10

  describe (Ok (Ok 7)), describe (Ok (Err "x")), describe (Err 4)
      interpreter: 7, -1, 40      wasm plug: 0, 0, 40      zig plug: refuses
```

The zig plug emitted, for `describe`, two bare `.Ok` prongs with `n` never bound:

```
return switch (v_) { .Ok => @compileError("zig plug: no emitter for n"), .Ok => (0 - 1), .Err => |code_| (code_ * 10),  };
```

In `ZigEmitter.codex`, `zig-pat-binder` answers `_` for any sub-pattern that is not a variable, and every path through `emit-zig-match-arm` and `emit-zig-chain-arm` reads a sub-pattern only through it. Separately, `zig-lit-pat-text` spells a Text literal pattern bare, so `is "hi" ->` at the top of an arm emits `hi` and does not build. In `WasmEmitter.codex`, `wat-ctor-sub-tests` tests a literal sub-pattern and sends a constructor sub-pattern to `is otherwise`; `emit-wat-bind-ctor-subs` binds nothing inside one. The first program is the silent case: a nested nullary constructor is a wrong answer with no diagnostic anywhere.

Found by two programs ported from the Roc language's test suite, in a corpus of 46 such ports we run on every backend. Neither the compiler's own source (the self-host is byte-identical through the zig plug) nor the 28 programs we cut from `codex/test` contains a nested constructor pattern that changes an answer.

## The fix

**zig.** A match with a nested pattern, or with a Text literal pattern at any level, is compiled through the chain form the emitter already uses for guarded matches: each inner test is an `if` inside the enclosing one, and a failed inner test falls out of the arm to the next, which a switch prong cannot do (no fall-through, and two prongs on one constructor are a duplicate). `emit-zig-chain-work` walks the sub-patterns as a work list of (pattern, value) pairs — a constructor sub-pattern replaces itself with its own sub-patterns at the head of the list, each addressed through the payload, with `.*` when the field's type is boxed — and emits the arm's tail once, when the list is empty. A Text literal compares through `cx_text_eq`. Every variable bound anywhere in the pattern is pushed into the context the arm's body is emitted with. A match with no nested pattern and no Text literal takes exactly the paths it took; a match with a literal sub-pattern took the switch path before, wrongly.

**wasm.** `wat-ctor-sub-tests-at` and `emit-wat-bind-ctor-subs-at` take the object's address as an expression. A constructor sub-pattern is a tag test on the object its payload field points at; its own bindings and tests are emitted inside that tag test, so a payload load never addresses past an object of the wrong tag. The last-arm shortcut (`wat-ctor-has-test-subs`, renamed from `-lit-`) treats a constructor sub-pattern as the test it is. The scrutinee's forms call the new ones at `(local.get scrut)`, so a pattern with no nesting emits what it emitted.

## Pinned by a test unit

`codex/test/ops/match-nested-pattern`: twenty lines — payload bindings two and three levels deep, a multi-payload constructor inside another, an Integer and a Text literal inside a constructor, a Text literal on its own, nested nullary constructors, and formulas that differ only inside the nesting and must fall through. Both plugs fail it before (zig refuses, wasm answers wrong) and print its `.expected` after; the interpreter prints it either way. The programs quoted above are the Roc ports, not this unit; the unit's `describe` arm is `is Ok (Err (t)) -> 0 - text-length t`.

## Measurement

Stack: our `u58-candidate` = Update 57 `49fa9f27` + PRs 135, 138, 139, 140 + this commit; the PR branch is off clean `49fa9f27` and was not built at that base (the zig plug at `49fa9f27` predates PR 135's boxing and does not self-host the compiler). Bare metal is not measured here beyond the fixed points.

- **zig, self-host, no guest** (`bootstrap_native.py`): the second self-host build emits the first's bytes, in one round; the compiler's own source emits byte-identical zig through the old and new emitter, so it contains no nested pattern and no Text literal pattern.
- **zig, fixed point under QEMU** (`build.py --force`): **HOLDS, byte-identical** (`build.py --force`, 530 s, three guests; `arith` matches all 9 lines).
- **wasm, fixed point** (`build.py --road both`): **HOLDS** by both roads (`build.py --road zig` then `--road both`; the module compiles its own source and gets itself back, and the native road agrees with it; the sample matches).
- **Three corpora we run by output** (28 cut from `codex/test`, 46 Roc ports, 54 specs of safari, a 54-chapter application of ours), through each plug rebuilt from this stack, before → after: zig 28/28 → 28/28, 42/46 → 44/46 (the two left are COMPILER-74, an unconstrained type variable the checker never defaults), 51/54 → 51/54 (safari's three are Real literals one ULP apart, issue 125, the plugs' parser); wasm 28/28 → 28/28, 44/46 → 46/46, 51/54 → 51/54. The two that move are the two Roc programs above.
- **wasm, corpus sweep** (`corpus_sweep.py` over the 1,269 `codex/test` programs at Update 57, IR from a hosted `codexir`, graded against their own `.expected`), the plug at `49fa9f27` versus the plug with this commit: 830 match, 152 trap, 26 differ, 4 refused — **the same numbers and the same 26 programs** on both sides. The wasm change moves nothing in the corpus.

Risks: both emitters, on the match arm only. Every match without a nested pattern takes the code path it took, which is what the corpus numbers above are checking.

### DIFF
--- a/codex/plugs/wasm/WasmEmitter.codex
+++ b/codex/plugs/wasm/WasmEmitter.codex
@@ -1317,24 +1317,27 @@ Section: Expression Emission
   emit-wat-ctor-pat (ctx) (name) (subs) (guard) (body) (branches) (i) (depth) =
    let tag = find-ctor-in-typedefs (ctx.type-defs) name 0
    in let bind-subs = emit-wat-bind-ctor-subs ctx subs 0
-   in if (i == list-length branches - 1) & wat-guard-trivial guard & (wat-ctor-has-lit-subs subs 0 == False) then
+   in if (i == list-length branches - 1) & wat-guard-trivial guard & (wat-ctor-has-test-subs subs 0 == False) then
     emit-wat-arm-block "" "(i32.const 1)" bind-subs (emit-wat-expr ctx body)
    else emit-wat-arm-block "" (wat-ctor-arm-cond ctx tag subs guard bind-subs) bind-subs (emit-wat-expr ctx body)
 
- A LITERAL inside a constructor pattern is a TEST, not a binding, and
- `emit-wat-bind-ctor-subs` had no arm for one: it bound a variable sub-pattern
- and sent everything else to `is otherwise`, so `BInt (0)` matched every `BInt`
- and every arm answered the first one's body. The tests fold onto the tag test
- rather than beside it, because a payload load on an object of the wrong tag can
- address past the object and trap, which is the reason the guard is already
- gated this way.
+ A sub-pattern of a constructor pattern is a binding (a variable), nothing
+ (a wildcard), or a TEST: a literal is an equality on the payload field, and
+ a constructor is an equality on the tag of the object the field points at,
+ followed by that constructor's own sub-patterns through that object. Every
+ test and every binding under a tag test sits INSIDE it, because a payload
+ load on an object of the wrong tag can address past the object and trap.
+ `wat-ctor-sub-tests-at` and `emit-wat-bind-ctor-subs-at` take the object's
+ address as an expression; the scrutinee's forms are the ones at
+ `(local.get scrut)`.
 
-  wat-ctor-has-lit-subs : List IRPat, Integer -> Boolean
-  wat-ctor-has-lit-subs (subs) (i) =
+  wat-ctor-has-test-subs : List IRPat, Integer -> Boolean
+  wat-ctor-has-test-subs (subs) (i) =
    if i >= list-length subs then False
    else when list-at subs i
     is IrLitPat (t) (ty) (sp) -> True
-    is otherwise -> wat-ctor-has-lit-subs subs (i + 1)
+    is IrCtorPat (n) (ss) (ty) (sp) -> True
+    is otherwise -> wat-ctor-has-test-subs subs (i + 1)
 
   wat-and-tests : Text, Text -> Text
   wat-and-tests (acc) (t) =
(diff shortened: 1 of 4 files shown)


damiant3:
Thank you, Steve. We are taking the nested-pattern fixes through the Zig and wasm owners, using one shared regression fixture. The Zig work follows PRs 135 and 138. Both owners will check failed inner tests falling through to later arms, nested bindings, text comparisons and wrong-tag payload access, with the reported failures reproduced before the fix.

Current Perforce main also carries newer decimal parsing, wasm exports and four-lane mask work. We will preserve those changes during ingestion. The PR stays open until the accepted work reaches a Git release; the closing update will identify the public commit and any changes made on top.

damiant3:
The wasm candidate passes your original 20-case fixture. The native check exposed a separate gap on our current seed: compiling that fixture reports eight CDX2040 errors for nested bindings (n, t, n, m, k, a, b, a), followed by CODEGEN-HALTED. That native path was outside the runtime measurements reported in this PR, so this does not contradict your interpreter/hosted results.

We are assessing the native pattern-emitter path before choosing the fixture's final placement. We will preserve the nested binding and fallthrough cases, rather than turn the unexpected refusal into the test's expected success condition. The Zig and expanded wasm checks continue; no action is needed from you yet.

damiant3:
Fixture disposition: we will keep the expanded regression under shared plug test inputs, with the successful output checked by the Zig and wasm harnesses. The native compiler's unresolved nested bindings remain an open defect with the original failing fixture and diagnostic preserved. We will not add an expected-refusal sidecar or describe the plug passes as native coverage.

The expanded wasm guard control also exposed eager guard evaluation after a failed nested match (integer divide by zero). Our ingestion includes a lazy guard-gating correction; the corrected wasm candidate passes all 31 expanded rows and the existing nested-guard, string-guard and constructor-equality regressions. We will call out that addition in the final acceptance and release record.

damiant3:
The wasm half is accepted and landed in Perforce as main 25578, with credit to Steve Howell. The three files are WasmEmitter and the shared fixture/expected pair under codex/plugs/test-input.

The final-path check passes all 31 expanded rows, the calibration rejects the sabotaged entry as intended, and the existing nested-guard, string-guard and constructor-equality regressions pass. Our addition is lazy guard evaluation after the nested pattern succeeds; the expanded control had otherwise trapped with integer divide by zero. Current exports, decimal parsing and four-lane masks are preserved. No runtime heap is added; nested address construction/load work scales with pattern nodes times depth.

The Zig half remains under review. Native nested-pattern support remains open as COMPILER-82, with the original eight CDX2040 failures preserved and no expected-refusal waiver. The PR stays open through completion and Git publication; the closing update will identify the public commit.

damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). Both halves are in: the wasm half in main
25578 (with lazy guard evaluation after a nested pattern succeeds, which the
expanded control needed) and the Zig half in main 25656 with PR 135. The
31-row shared fixture lives in `codex/plugs/test-input`. The native x86-64
backend has the same nested-pattern gap; it is open as COMPILER-82 on our
side and is not part of this PR. Thank you.

PR #142, opened 2026-09-11
Parser: a class or instance declaration begins its line (COMPILER-81)

showell:
`scan-class-instance-defs` walks the whole token stream, and the lexer keeps a section title's words as tokens, so `instance` or `class` followed by a type identifier inside a title declares a phantom instance or class. Two valid programs are refused:

- `Section: Notes (see instance Showable Text below)` gives `Showable` a second instance; the desugarer builds `Showable-dict-Text` from no methods and the checker refuses with CDX2006.
- `Section: More notes (the class Showable where it matters)` declares the class twice; the checker refuses with CDX3001 on the duplicate dictionary type.

The scan now takes the keyword only when it begins its line (`begins-line`: the first token, or one whose predecessor is a newline). Every class and instance declaration in the tree sits there: 26 at column one, four indented by two. A title's words never begin a line.

Pinned by `codex/test/ops/section-title-keywords`, a valid program carrying both titles, expected `<42>`. Registered as COMPILER-81 in the compiler backlog.

**How it was found and what was verified.** An independent front end (rust-codex-compiler, Steve Howell's) implements the same class scan at item starts only; its checker minted two fewer type variables than `codexcheck` on `nrf52840-drivers`, whose one section title reads "shared with SPIM via instance ID". That front end's corpus gates agree with `codexcheck` on all 1,269 corpus units (diagnostics: count and first code) with this rule in place. Under QEMU the seed compiles the patched compiler source to IR cleanly (3m09s). The pinning unit is refused by the unpatched `codexcheck` at `8570fba1` and runs to `<42>` through the independent front end.

**What was not verified, and why.** A native build of the patched `codexir` on this box did not complete: the zig plug at Update 58 refuses the compiler's own emitted zig on two points unrelated to this change, a local `cx_new` shadowing the prelude's `cx_new` in `zigemit.zig`, and `zig plug: no address-of for codexir.IRActStmt` / `IRPat` in `codexir.zig`. Both reproduce on clean `5b3e14a7`. They look like Update 58 plug gaps worth their own rows; happy to file them if useful. So the corpus census through a patched native `codexir` is owed, and the claim here rests on the seed compile plus the independent front end.

Written by Claude (Anthropic) working with Steve Howell; risks stated above.


damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4), seed 49070BAEB1E31085. Landed as main 25660 with a new
signed seed, credited to you. `scan-class-instance-defs` takes `class` and
`instance` only at the start of a line. Your fixture
`codex/test/ops/section-title-keywords` is refused with CDX3001 by the old seed
and prints `<42>` on the new one; the existing typeclass tests, the
self-compile fixed point and the BVT (143 of 143) pass. One change from the
PR: COMPILER-81 was already taken in our backlog, so the row is COMPILER-84.
The two Zig refusals you noted (`cx_new` shadow, `IRActStmt`/`IRPat`
address-of) are what the `cx_frac` rename and PR 135's boxing address, both in
this update; we have not rebuilt a native `codexir` through Zig to confirm it.
Thank you.

PR #143, opened 2026-09-12
zig plug: cx_real_to_text remainder must not be named cx_new

apoorvapendse:
The zig prelude's `cx_real_to_text` used a local named `cx_new` for the fractional remainder of a Real. That is also the heap-box helper (`fn cx_new`). Zig refuses a local that shadows a file-scope declaration, so `zig build-exe` dies on any subject that pulls in real-to-text, including `zigemit`.

`fib` never prints a Real, so it built. `zigemit` did not:

```
zigemit.zig:3891:9: error: local variable shadows declaration of 'cx_new'
    var cx_new = cx_e - cx_acc;
zigemit.zig:4422:1: note: declared here
fn cx_new(v: anytype) *@TypeOf(v) {
```

The reserved list already rewrites *user* identifiers away from `cx_new`. Prelude-internal locals are handwritten strings and never go through that pass.

**Fix.** Rename the local to `cx_frac` and reserve that name on `zig-prelude-decls`. `fn cx_new` is unchanged.

**Verified** on this branch (`b8a072f8`) through [cobblestone-qemu](https://github.com/showell/cobblestone-qemu) `./build.sh zigemit` against the seed: ring plug 30s, compile 30s, transpile 14s, native `zigemit` 2,769,784 bytes. The emitted helper now has `var cx_frac` at the old site and `fn cx_new` still at the boxer. No leftover guests.

This is the `cx_new` shadow Steve noted as an unrelated Update 58 plug gap on PR 142.

Not verified: official `sweep.ps1` / pingpong / BVT (those need the Windows box). The change is a prelude local rename; it does not touch IR or the seed.


damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). The same rename was already carried by Steve
Howell's PR 135 intake (main 25656): the prelude local is `cx_frac` and the
name is reserved, byte for byte what this PR proposes, so no separate change
landed. Thank you for the clear diagnosis and the zigemit reproduction.

PR #144, opened 2026-09-12
arm64 plug: Darwin userspace runtime and Mach-O wrap

apoorvapendse:
The ARM64 plug's virt path writes the PL011 UART and boots as a QEMU kernel. This adds a Darwin userspace arm: mode `IR-CCE darwin` emits `mmap` / `write` / `exit` via `svc #0x80`, and a host wrap turns the wire into a signed arm64 Mach-O.

**Runtime.** Slab at 8 GiB (`MAP_FIXED|MAP_PRIVATE|MAP_ANON`), above the 4 GiB PAGEZERO. CCE tables at slab+0x1000 / +0x1100, heap at slab+0x2000. `x28` is the bump pointer, so print uses the fixed table addresses, not `x28`. Virt UART / EL1 start is unchanged.

**Host wrap.** `wrap_macho.py` concatenates code + 8-byte-padded data (same layout `a64-patch-rodata` assumes), trampolines `_start` to `__start`, links with `ld -lSystem`, ad-hoc codesigns. `compile-macho.py` is the Mac pipeline: seed IR, plug, wrap. `run.ps1 -Darwin` passes the mode on the Windows serial path.

**Also.** Prologue/epilogue pair `x27` with `xzr`. `ldp x27, x27` is constrained unpredictable and SIGILL on Apple Silicon. QEMU virt accepted it, so the virt battery never saw it.

**Verified** on an M5 host, this commit, against seed `49fa9f27`. `eight-queens.codex` through `compile-macho.py` produced `~/runs/20260912-arm64-darwin/queens` and printed the first board, exit 0. Same board as the seed-compiled CDX booted under QEMU. Plug compile ~100s; later runs reuse the sandbox CDX.

Not verified: official `sweep.ps1` / BVT / Renode virt regression. The virt emit path is the same except the x27 pair, which is a no-op on QEMU.

Does not include the demo `eight-queens.codex` (left untracked).

### DIFF
--- a/codex/plugs/arm64/Arm64CodeGen.codex
+++ b/codex/plugs/arm64/Arm64CodeGen.codex
@@ -114,6 +114,7 @@ Section: Codegen State
    wcet-reports : List Text,
    state-local : Integer,
    smp-boot : Boolean,
+   darwin-boot : Boolean,
    runtime-funcs : Integer,
    boot-cap-mask : Integer,
    boot-fs-scope : Text,
@@ -224,6 +225,7 @@ Section: State Constructors
      wcet-reports = [],
      state-local = 0 - 1,
      smp-boot = False,
+     darwin-boot = False,
      runtime-funcs = 0,
      boot-cap-mask = 0,
      boot-fs-scope = "",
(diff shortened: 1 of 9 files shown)


damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). The Codex half landed as main 25665, credited
to you: the `IR-CCE darwin` mode (Darwin syscalls through `svc #0x80`, the
8 GiB slab, `WIRE-END`), the `stp`/`ldp x27, xzr` fix, and `run.ps1 -Darwin`.
Two adjustments: the virt path's serial output stays byte-identical (the PR
added a second blank line after every virt wire), and two prose paragraphs
were trimmed. The Python wrap scripts and `bundle-qemu.ps1` did not enter the
tree: the build takes no dependency outside Windows and codex-vm, so the
Mach-O wrap lives in your repository (recorded in plugs-backlog 2.57). We
checked the virt path by building the plug before and after and diffing the
wire for six programs: only the `x27` pair bytes differ, and five programs
including one with seven changed pairs pass on the arm64 Renode board. Nothing
in our tree
runs a Darwin binary, so your M5 run is the only execution on record. Thank
you.

PR #145, opened 2026-09-12
examples: eight-queens prints the first board

apoorvapendse:
A self-contained 8-queens program. It searches row by row, columns left to right, and prints the first board:

```
Q . . . . . . .
. . . . Q . . .
. . . . . . . Q
. . . . . Q . .
. . Q . . . . .
. . . . . . Q .
. Q . . . . . .
. . . Q . . . .
```

That is columns `[0, 4, 7, 5, 2, 6, 1, 3]`. Placement lists are copied on each push so `list-push` does not alias under backtracking. No Foreword cites.

Lives next to the other sample under `codex/test/examples/`, with an `.expected` sidecar.

**Verified** on Update 57 `49fa9f27`: seed-compiled CDX under QEMU, and the same source through the Darwin ARM64 Mach-O path (PR 144). Both printed this board.

Independent of PR 143 and PR 144.


damiant3:
Published in Update 59, [3fefc985](https://github.com/damiant3/Cobblestone/commit/3fefc98557923071f9266c624ea44de8f0bc48f4). Landed as main 25663 in
`codex/test/examples`, where the battery picks it up; the output matches your
expected board on the release seed. We dropped the one-line prose summary
per our source rule. Thank you.

PR #148, opened 2026-09-13
fat16 test fixtures: the second FAT matches the first

showell:
direction.*

Branch: showell/NewRepository `fat16-fixtures-fat2-reserved`, one commit on
Update 60 (`9fff850c`).

## What changes

The eleven `codex/test/fat16-*.disk` fixtures change, bytes only; no source
changes. In each fixture the second FAT now equals the first.

- **Reserved entries, all eleven.** The second FAT held zeros in entries 0
  and 1, where the first holds `0xFFF8` and `0xFFFF`. Four bytes change per
  file, at offset 1,110,528: the ESP at sector 2048, one reserved sector, then
  120 FAT sectors. `00 00 00 00` becomes `f8 ff ff ff`.
- **The cycle, `fat16-cycle-guard.disk` only.** Its `2 -> 3 -> 2` cycle was
  in the first FAT only; the second held `0xFFFF` for both clusters. The
  test's prose (line 7) says "in both FAT copies", and so does
  `docs/ExaminersAssay.md` (line 4169). Four more bytes, at 1,110,532, make
  that true.

## Why

A fixture should be a valid FAT volume apart from the defect it was built to
carry. `fsck.fat` (dosfstools 4.2) on the untouched ESPs reports
`FATs differ` for all eleven, before any test runs.

**The reserved entries follow the image builder's history.**
- At `faf1c639^`, `build/build-img.ps1` wrote the reserved entries to the
  first FAT only (`W16 $fat1Off 0xFFF8` and `W16 ($fat1Off + 2) 0xFFFF`,
  lines 169-170), while it wrote cluster chains to both copies.
- `faf1c639` (Update 35) brought in `Set-Fat`, which writes every entry to
  both copies.
- Six of the fixtures (`alloc`, `dirgrow`, `list`, `mkdir`, `subdir`,
  `write`) first appear under `codex/test/` in that same commit, and the rest
  arrive later. `fat16-longname.disk` was added in `58b08c38`, on 2026-08-28.
  All eleven carry the first-FAT-only form.
- That the builder made them is an inference from their layout, not traced:
  a GPT image whose FAT16 ESP holds `CODEX.CDX` and `EFI/BOOT/BOOTX64.EFI`.

## Verified

- **The bytes.** `cmp -l` against Update 60 shows four changed bytes per
  fixture (eight in `fat16-cycle-guard.disk`) at the offsets above. After the
  patch the two FATs are identical across all 120 sectors in every file.
- **fsck.fat 4.2, before and after.** Before, all eleven report
  `FATs differ`. After, ten pass cleanly. `fat16-cycle-guard` still fails,
  with only the circular chain under `/EFI` and the messages that follow
  from it.
- **No verdict reads the changed entries**, by reading the source.
  - `fat16-cluster-ok` rejects clusters below 2, and the free scan starts
    at 2.
  - `fat16-next-cluster` reads the first FAT, and that is how
    `fat16-cycle-guard` walks its chain.
  - The only test that reads the second FAT is `fat16-alloc`
    (`fat16-read-fat-entry vol 1 free1` and `vol 1 got`), and it reads only
    the clusters it allocates.
- **The one write path that can reach them.** `fat16-write-fat-copies`
  reads, patches and writes back a sector of each copy. A write whose entry
  falls in the second FAT's first sector (clusters below 256) now carries the
  reserved entries through as `0xFFF8`/`0xFFFF` instead of zeros. No fixture
  has a free cluster below 256, so no allocation lands there. Freeing an
  existing low chain could; that was not checked.
- **By run, outside the battery.** An independent port of Foreword `Fat16`,
  run over a model of codex-vm's IDE disk, gives the same result for all
  eleven programs on the patched and original fixtures. The Cobblestone
  battery and codex-vm were not run here.

**To reproduce:**

    # patch: make the second FAT equal the first
    python3 -c "import struct,sys
    for f in sys.argv[1:]:
        b=bytearray(open(f,'rb').read()); p=2048*512
        bps,res,fs=struct.unpack_from('<H',b,p+11)[0],struct.unpack_from('<H',b,p+14)[0],struct.unpack_from('<H',b,p+22)[0]
        f1=p+res*bps; b[f1+fs*bps:f1+2*fs*bps]=b[f1:f1+fs*bps]; open(f,'wb').write(b)" codex/test/fat16-*.disk
    # check one ESP
    dd if=codex/test/fat16-list.disk of=esp.img bs=512 skip=2048 && fsck.fat -n esp.img

## Risks

- **A new hash.** A tool that hashes a fixture file whole sees a new hash.
  `git grep` finds no hash or manifest that names these files, and
  `bvt.ps1` copies each fixture to a temporary file.
- **The cycle may belong in the first FAT only.** If so, it is the prose
  that should change, and the `fat16-cycle-guard.disk` part of this commit
  should be dropped.


damiant3:

Your FAT16 fixture repair landed in Perforce CL 25746. All eleven images match your Git blobs exactly, with both FAT copies consistent and the intended cycle preserved.

Issue #153, opened 2026-09-22
An empty list literal type-checks as any type: `v : Integer = []` compiles clean


showell:
Update 61 (`3af67f99`, seed `7BCD5BC6BCE0AF41`). Also present at Update 60.

The checker gives an empty list literal a bare fresh type variable rather than
a list type, so `[]` unifies with `Integer`, with `Text`, or with anything
else, and the program compiles with no diagnostics. The same looseness puts
`error` on the wire in a clean compile of a nested empty literal (`[[]]`).
COMPILER-81's repair in Update 61 is correct on its own terms; here it closes
a variable that was only unconstrained because of this, which makes the IR
say `Integer` where there is a list.

## Reproducers

Each is a whole unit, cites resolved as `build/compile.ps1` resolves them:

```
Chapter: M
  cites Foreword chapter Console

Section: F

  v : Integer
  v = []

  opening : [Console] Nothing
  opening = act
   print-line-uni (show v)
  end
```

and the same unit with `v = text-length []`, and a control with `v = [1] + 1`.

## What we measured

The seed on bare metal (QEMU), mode `IR-CCE`, Update 61:

| `v =` | diagnostics | IR for `v` |
|---|---|---|
| `[]` | none | `(def "v" "M" (params) int-default (list-expr (elems) error) 0 0)` |
| `text-length []` | none | `(apply (name "text-length" (fn text int-default)) (list-expr (elems) error) int-default)` |
| `[1] + 1` (control) | CDX2001 `List vs Integer`, CDX2003 `Arithmetic operator requires Integer or Real`, CDX2001 `Integer vs List` | none emitted |

So `[]` is accepted as an `Integer` in the first and as a `Text` in the second.

At Update 60 we checked with the checkout's compiler built as a native program
through the Zig plug, stopped at the IR, whose bag halts emission the way the
driver's does: the first two emit the same IR for `v`, and the control halts on
the same CDX2001. (At Update 61 that native build emits the seed's IR byte for
byte on these units.)

We have not run either program; the claim is about what the checker accepts.

## Where it comes from

`Types/TypeCheckerInference.codex:1384-1391`, `infer-list` on zero elements:

```
   in if list-length elems == 0
    then
     let (fr-ty, fr-st) = fresh-and-advance st
     in CheckResult {
      inferred-type = fr-ty,
```

The literal's type is `fr-ty` itself, not `ListTy fr-ty`, so nothing constrains
[...]

PR #154, opened 2026-09-22
Two refusal tests for #153: an empty list literal where an Integer or a Text is wanted

showell:
Two refusal tests for #153, as asked, and a backlog row for the gap.

**These tests are red at Update 61 by design.** They pass when the checker stops accepting `[]` as a non-list; nothing in this PR changes the compiler.

| test | source | `.failing` |
|---|---|---|
| `codex/test/errors/empty-list-as-integer` | `opening : Integer = []` | `2001` |
| `codex/test/errors/empty-list-as-text` | `opening : Integer = text-length []` | `2001` |

## Why 2001

The same two programs with `[1]` in place of `[]` are refused with exactly CDX2001 and nothing else. Measured on the Update 61 seed (`3af67f99`, `7BCD5BC6BCE0AF41`) on bare metal (QEMU), mode `IR-CCE`, with the unit `build/compile.ps1` would assemble:

| `opening : Integer =` | diagnostics |
|---|---|
| `[1]` | CDX2001 |
| `[]` | none |
| `text-length [1]` | CDX2001 |
| `text-length []` | none |

The two test files in this branch were compiled the same way and both compile clean today, so both will fail in the battery until the fix. If the eventual repair reports a different or an additional code, the sidecars are the thing to change; the claim they carry is only that these programs must be refused.

## The backlog row

COMPILER-85, at the top of the table in `codex/compiler/compiler-backlog.md`. It names the site (`infer-list`), the two tests, and #153. Renumber it if 85 is taken on your side.

Not included: the nested shape from #153 (`[[]]` carrying `error` in a clean compile) and the `MkTup2` typing in the issue's comment. Neither is a refusal, so neither fits `errors/`; `build/ir-fidelity` looks like their home, and we have not run it here.

---


damiant3:

Your two refusal tests landed unchanged, with their `.failing` sidecars, in Perforce main CL 26013, together with the fix for #153: an empty list literal is now a `List` or a `LinkedList` and nothing else, so `v : Integer = []` and the Text case are refused with CDX2001. The second shape from #153 (desugared tuples, revised records and derived `show` losing their instantiated IR types) was fixed as COMPILER-85 in main CL 26027; your `apply_check.py` over your 27 programs went from 584 contradicted sites to 0.

Closing with that commit.

PR #155, opened 2026-09-22
Zig plug: emit bit-not

showell:
The Zig plug has no emitter for `bit-not`, a declared builtin (`Types/Builtins.codex`) that the wasm plug already emits as `(i64.xor x (i64.const -1))`. Any program that reaches SHA-512 (`Foreword Sha512`) refuses to build on the Zig plug with `zig plug: no emitter for bit-not`, which since Update 61's crypto tests includes Ed25519, X25519 and ECDSA P-384.

The change is one entry in the builtin emitter table, beside `bit-and`, `bit-or` and `bit-xor`:

```
ZigBuiltinEmitter { name = "bit-not", emit = \args ctx d ty -> "(~@as(i64, " & ... & "))" },
```

The `@as(i64, ...)` is there because a literal argument reaches zig as a `comptime_int`, which has no width for `~`; every `IntegerTy` is `i64` in this plug (`emit-zig-type`), so the pin changes nothing else.

## The test

The battery already has one: `codex/test/tco-bitop-loop` exercises `bit-not` and passes on bare metal. What it could not do was build through the Zig plug. Measured with the Zig plug built from this branch (the Update 61 transpiler transpiling this branch's plug source natively, then `zig build-exe`), against each test's own `.expected`:

| program | Update 61 | this branch |
|---|---|---|
| `tco-bitop-loop` | build refused: no emitter for bit-not | match |
| `apps/x25519-vector-test` | refused | match |
| `apps/ed25519-sign-test` | refused | match |
| `ecdsa-p384` | refused | match |
| `apps/cdx-binary-test` | refused | match |

`e1000-ulp-entry` also uses `bit-not`; with this change it gets past it and stops at `port-out-32`, a device builtin the Zig plug does not emit, which is expected.

The prelude surface is unchanged (no new prelude part), so `check-zig-prelude-surface` has nothing new to grade.

---


damiant3:

The `bit-not` emitter landed as your PR head, unchanged, in Perforce main CL 26007 (the PR's base blob equalled the depot file). One addition on top: the stale "Integer arithmetic wraps" prose in `ZigEmitter.codex` that #157 pointed out was deleted in the same change. Proof: with the old plug, `tco-bitop-loop` and three crypto subjects were refused by `zig run`; with yours all four match their expected output.

Closing with that commit.

